PR #4, opened 2020-05-17
Fix Shop/Upgrades text colors

dengr1065:
Right now text in Upgrades dialog doesn't look ok while using dark theme, this PR fixes it. Although code style might not be the same, this will fix text appearance for now.


tobspr:
Thanks!

PR #6, opened 2020-05-20
Add Item Buffers

jmeggitt:


src/js/game/components/item_processor.js
```diff
@@ -30,7 +31,7 @@ export class ItemProcessorComponent extends Component {
             nextOutputSlot: types.uint,
             type: types.enum(enumItemProcessorTypes),
             inputsPerCharge: types.uint,
-
+            chargeWhenBlocked: types.bool,
```
jmeggitt, line 34: This change may break save data. It may be worth considering reverting it and having the buffer use 0 for `inputsPerCharge`.

src/js/game/key_action_mapper.js
```diff
@@ -44,6 +44,7 @@ export const KEYMAPPINGS = {
         mixer: { keyCode: key("8") },
         painter: { keyCode: key("9") },
         trash: { keyCode: key("0") },
+        buffer: { keyCode: key("-") },
```
jmeggitt, line 47: This may not be a valid key, it likely needs to be changed.

src/js/game/systems/item_buffer.js
```diff
@@ -0,0 +1,52 @@
...
+
+    draw(parameters) {
+        this.forEachMatchingEntityOnScreen(parameters, this.drawEntity.bind(this));
+    }
+
+    /**
+     * @param {DrawParameters} parameters
+     * @param {Entity} entity
+     */
+    drawEntity(parameters, entity) {
```
jmeggitt, line 23: This implementation does not take into account the building rotation.

jmeggitt:
Another issue that needs to be resolved before this can be used is that it does not validate that the items it accepts have the same definition as the storage.

jmeggitt:
Added with 2161aa6.

PR #9, opened 2020-05-21
Allow deselect through toolbar

jmeggitt:
When I first tried playing the game I found it annoying how I couldn't deselect a building by pressing on its icon in the toolbar. This small tweak that allows users, stop building placement by simply pressing on the toolbar icon or hotkey a second time.

This also includes a slight update to the key action mapper that changes its behavior from sending a signal periodically when a key is held to only sending a signal on the initial key stroke. This was done to prevent flickering if a building hotkey is held down. I was unable to find any instances where this change negatively affected other parts of the game.

### DIFF
--- a/src/js/game/hud/parts/buildings_toolbar.js
+++ b/src/js/game/hud/parts/buildings_toolbar.js
@@ -143,8 +143,25 @@ export class HUDBuildingsToolbar extends BaseHUDPart {
             return;
         }
 
-        this.root.soundProxy.playUiClick();
-        this.sigBuildingSelected.dispatch(metaBuilding);
-        this.onSelectedPlacementBuildingChanged(metaBuilding);
+        let previouslySelected = false;
+        for (const buildingId in this.buildingHandles) {
+            const handle = this.buildingHandles[buildingId];
+            if (handle.metaBuilding === metaBuilding) {
+                if (handle.index === this.lastSelectedIndex) {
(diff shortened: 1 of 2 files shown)


src/js/game/hud/parts/buildings_toolbar.js
```diff
@@ -143,8 +143,25 @@ export class HUDBuildingsToolbar extends BaseHUDPart {
...
+                    previouslySelected = true;
+                }
+                break;
+            }
+        }
+
+        const buildingPlacer = this.root.hud.parts.buildingPlacer;
+
+        if (previouslySelected && buildingPlacer.currentMetaBuilding.get()) {
+            buildingPlacer.abortPlacement();
```
tobspr, line 160: There should be a .playUiClick here as well I think, but I will simply add this afterwards

tobspr, line 160: Instead of using abortPlacement directly, simply use `this.sigBuildingSelected.dispatch(null)` - The idea is that the HUD parts only interact via events and do not refer to other parts explicitely (I know I did break that myself at some points)

tobspr: I would refactor this a bit and instead of having a `previouslySelected` variable, simply set `metaBuilding = null` in line 151 - Then you don't need the whole if construct

PR #10, opened 2020-05-23
Basic Waypoints

dengr1065:
This PR adds simple waypoints.

Themes: light, dark
UI Scale: super small - huge
You can add and remove waypoints, teleporting closes the window. The name field also acts as search, this is helpful if you have lots of waypoints.

This feature should be available in demo version as well, without any ads in its UI. And even if you think that it should be overlay-only, it should not be that. With dark theme it looks just cool and well-integrated with other UI, if it is an overlay, it's not that cool. Overlay *might exist* as well, but just as marker/pointer.

Further updates:

- Add overlay to show existing waypoints (**just to show, please**) that can be disabled in settings.
- Make it consistent with other engine methods

alt


src/css/ingame_hud/waypoints.scss
```diff
@@ -0,0 +1,95 @@
+#ingame_HUD_Waypoints {
+    .content {
+        @include S(width, 500px);
+    }
+
+    .wizardWrap {
```
Playez, line 6: cul wizard

mse-k:
since its implementd close this xd

PR #11, opened 2020-05-24
Add unstacker building

jmeggitt:
Adds the framework for a new unstacker building as a variant to the existing stacker. A level goal still needs to be added for unlocking this variant and I did not make an asset for it since I cannot easily recreate the blueprint aesthetic in inkscape.

General concept for addition:
Capture


src/js/game/shape_definition.js
```diff
@@ -465,6 +465,43 @@ export class ShapeDefinition extends BasicSerializableObject {
...
+        }
+
+        // Check if there is only one layer
+        if (this.layers.length === 1) {
+            return [new ShapeDefinition({ layers: lowerLayers })];
+        }
+
+        let upperLayers = [];
+
+        layerCheck:
```
tobspr, line 485: This code can be vastly simplified - simply always split it into two shapes, the first one containing `layers[0 ... layers.length - 2]` and the second one containing just `layers[layers.length-1]`.

In the item processor, then emit both shapes, but for each shape check if `definition.isEntirelyEmpty()` and if so, don't emit anything. This will also make it more consistent

src/js/game/shape_definition_manager.js
```diff
@@ -144,6 +144,21 @@ export class ShapeDefinitionManager extends BasicSerializableObject {
...
+     * @returns {[ShapeDefinition, ShapeDefinition]}
+     */
+    shapeActionUnstack(definition) {
+        const key = "unstack:" + definition.getHash();
+        if (this.operationCache[key]) {
+            return /** @type {[ShapeDefinition, ShapeDefinition]} */ (this.operationCache[key]);
+        }
+
+        const unstacked = definition.cloneAndUnstack();
+        return /** @type {[ShapeDefinition, ShapeDefinition]} */ (this.operationCache[key] = unstacked);
```
tobspr, line 159: This is not correct - it sometimes just returns `[ShapeDefinition]`. If you adjust it the way I commented above, this comment is obsolete though.

src/js/game/systems/item_processor.js
```diff
@@ -23,7 +23,7 @@ export class ItemProcessorSystem extends GameSystemWithFilter {
             // First of all, process the current recipe
             processorComp.secondsUntilEject = Math_max(
                 0,
-                processorComp.secondsUntilEject - this.root.dynamicTickrate.deltaSeconds
+                processorComp.secondsUntilEject - this.root.dynamicTickrate.deltaSeconds,
```
tobspr, line 26: I think your IDE added a trailing comma there, however the prettier config has this explicitely disabled. Not a big issue, I can reformat the file after merging it.

src/js/game/systems/item_processor.js
```diff
@@ -196,14 +196,31 @@ export class ItemProcessorSystem extends GameSystemWithFilter {
...
+            // UNSTACKER
+
+            case enumItemProcessorTypes.unstacker: {
+                const item = /** @type {ShapeItem} */ (itemsBySlot[0].item);
+                assert(item instanceof ShapeItem, "Input for unstack is not a shape");
+
+                const unstackedDefinitions = this.root.shapeDefinitionMgr.shapeActionUnstack(item.definition);
+
+                for (let i = 0; i < unstackedDefinitions.length; ++i) {
+                    outItems.push({
```
tobspr, line 216: As mentioned above - use `isEntirelyEmpty()` here and if so, don't emit on this slot

src/js/game/tutorial_goals.js
```diff
@@ -11,6 +11,7 @@ export const enumHubGoalRewards = {
     reward_painter: "reward_painter",
     reward_mixer: "reward_mixer",
     reward_stacker: "reward_stacker",
+    reward_unstacker: "reward_unstacker",
```
tobspr, line 14: If you add a reward, there also has to be a translation for it. I can add it tho

tobspr:
Code looks fine so far. Would have to take a closer look. But first I need to figure out where this building should be used, and I probably would also introduce some new mechanics. So, this one will have to wait a bit

tobspr:
I'll close it for now since  I don't have plans for merging it in the near future. Its not lost and can be serve as reference for other buildings

PR #12, opened 2020-05-25
Consistency update to upgrade text

CallMeHein:
Changed "Shape Processing" to "Cutting, Rotating & Stacking" to stay consistent with the other upgrades which also list which building they affect.


tobspr:
Thanks! I agree that this makes more sense

PR #19, opened 2020-05-26
Git LFS check for gulpfile

dengr1065:
Simple check for Git LFS in gulpfile to prevent confusion, also some readme changes.


tobspr:
Nice, thanks!

PR #23, opened 2020-05-27
Make ctrl/shift/alt keubindings changeable

Dimava:
Also consider changing 'no-rotate' default to shift


tobspr:
Thanks a lot! I'll consider it, but changing it now might confuse people

PR #24, opened 2020-05-27
Make CycleBuildings work properly

Dimava:
Make CycleBuildings work properly
 - ignore locked (instead of: raise error sound)
 - select last used if empty (instead of: select next)


tobspr:
Thanks!

PR #25, opened 2020-05-27
gitignore update

ModdedGamers:
Updates .gitignore so git push won't complain about workflow files on cli.


tobspr:
I don't think thats a good idea - Why do you want to add it to the gitignore?

ModdedGamers:
Because if you have a Github api token, and it doesn't have the workflows permission, You can't push without removing the file from git anyways. This just makes it easier to push without having to remember to remove the file. Plus, people will almost always work on the workflow files on the github website. So, not tracking these on remote machines is a better idea to create less hassle.

tobspr:
But shouldn't you be able to just work in your local repository?

ModdedGamers:
Even if you don't change the file, github won't let you push without explicitly removing it from git's index.

tobspr:
Thats weird, but I'll just accept it then I guess

PR #26, opened 2020-05-27
Add swapper building

hexagonhexagon:
Add a building that swaps the two input belts.

Things I haven't done:
 - Include the texture packer atlas. I don't have a license for it, so I can't do it. All the images are already there, you just have to pack them.
 - Add a new milestone for the swapper. I wasn't sure where to put it/what it should cost, so I thought I'd let you weigh in on where this should go.

Feel free to improve the art, I had a tough time working with it because I don't have Photoshop, and didn't want to trace it in Inkscape.


tobspr:
PR looks good, but I can't merge it without figuring how this would fit into the gameplay

hexagonhexagon:
Sorry I didn't respond earlier, wanted to play the new update and all that. Here are 2 examples of the swapper where it makes something significantly easier than it was before.

1. This example where I can alternately swap in two different things, paint them 2 different colors (not belted in here), and merge them. This would take significantly more space without the swapper.
Screenshot from 2020-05-29 18-25-29

1. This example where I have a bunch of belts that are collated, and I can rearrange them easily with the swapper to put into a white dye maker. No need to do a ton of nonsense with undergrounds in order to make the things go the right places.
Screenshot from 2020-05-29 18-27-33

hexagonhexagon:
You haven't commented on this PR for a while now. If you don't plan on adding it, can you close this pull request?

tobspr:
Sure, yeah well I wanted to keep it open because I feel bad if the work is lost. But RN I don't feel like the game does need this building

dengr1065:
i feel like this building should be present in wires update, and not require energy (a variant of balancer, unlocked at level ~21)

RogiJAG:
> i feel like this building should be present in wires update, and not require energy (a variant of balancer, unlocked at level ~21)

This could fit in with the idea we've been talking about where you unlock rewards through upgrades, but that hasn't been decided on yet.

PR #30, opened 2020-05-29
Add Sorters

mse-k:
adds sorters, they put the first item they get inputted out tyhe front and evey item after that is compared to the original and if the same will also go to the front BUT if not will go to the right

Needed stuff: Sprites, Description, New color that fits the theme better, Level to unlock and a goal to do so, make it a variant of belt, make it output evenly to both sides instead of just to 1 for the doesnt match side

Edit: savegames DO work


mse-k:
ive noticed a chek has failed but atlest for me the whle thing compiles and works just fine so idk

tobspr:
1. Please use meaningful commit names. It doesn't have to be an essay, but "aaaaaaa" is not a good commit name.
2. I'm not sure how sorters would fit into the current game play. Need to figure this out first

mse-k:
> 1. Please use meaningful commit names. It doesn't have to be an essay, but "aaaaaaa" is not a good commit name.

yea ik i usaly wouldnt name it that way but i was mostly just saving it in vs code so it just was like "push commit to master" and i just spammed aaaaaaa in bc i was lasy and wanted to continue

and plus it doesnt really matter since all i was doing was continuing implementation (back then it didnt even load and im pretty sure i couldnt even compile yet)

tobspr:
I'll close this PR since I have no plan for adding this building in the near future

PR #33, opened 2020-05-29
v1.1.1 french translation

TheCorsaireBe:


tobspr:
Nice, thanks! Can't give an estimate when I will add support but I'm trying to do it quickly

PR #41, opened 2020-05-29
Add linux build to standalone building process

hexagonhexagon:
Add a linux build. Not much more to it than that.


tobspr:
Have you tried it on a linux system? Since for me it doesn't seem to work

hexagonhexagon:
I use this branch to build standalone for Linux on my Ubuntu machine, so it definitely works for me. What do you mean when you say it doesn't work for you?

tobspr:
dengr tested it, but I guess I should offer experimental mac osx and linux builds

tobspr:
I just added experimental builds

PR #43, opened 2020-05-29
Better big number formatting

hexagonhexagon:
I redid the formatBigNumber() function to include a decimal point, and it also supports numbers up to 999.9T (although it isn't really necessary). Results below: 
Screenshot from 2020-05-29 18-49-10


tobspr:
LGTM! But could you move those suffixes to `base-en.yaml` so they can be translated?

hexagonhexagon:
I have done so now: they are under the key "global.suffix".

tobspr:
Thanks a lot!

PR #44, opened 2020-05-29
Make upgrade formatting more clear

hexagonhexagon:
Change the "Speed +x%" to "Speed x1 -> x2", like so:
Screenshot from 2020-05-29 19-26-33
Screenshot from 2020-05-29 19-26-53

### DIFF
--- a/translations/base-en.yaml
+++ b/translations/base-en.yaml
@@ -288,7 +288,7 @@ ingame:
         # The roman number for each tier
         tierLabels: [I, II, III, IV, V, VI, VII, VIII, IX, X]
 
-        maximumLevel: Maximum level
+        maximumLevel: MAXIMUM LEVEL (Speed x<currentMult>)
 
     # The "Statistics" window
     statistics:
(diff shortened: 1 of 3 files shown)


translations/base-en.yaml
```diff
@@ -264,7 +264,7 @@ ingame:
         # The roman number for each tier
         tierLabels: [I, II, III, IV, V, VI, VII, VIII, IX, X]
 
-        maximumLevel: Maximum level
+        maximumLevel: MAXIMUM LEVEL (Speed x<currentMult>)
```
tobspr, line 267: Could you make this lowercase, e.g. `Maximum Level`? I prefer to have everything properly cased and then do `text-transform: uppercase`

translations/base-en.yaml
```diff
@@ -316,16 +316,16 @@ ingame:
 shopUpgrades:
     belt:
         name: Belts, Distributor & Tunnels
-        description: Speed +<gain>%
+        description: Speed x<currentMult> → x<newMult>
```
tobspr, line 319: I feel like `Speed x1 -> x2` is a bit unclear / doesn't look that nice, maybe we can find a better wording?

translations/base-en.yaml
hexagonhexagon: The reason I did this was because if I left text-transform: uppercase as is, I couldn't have "Speed x15" show up as lower case, as it would show as "SPEED X15", which I thought didn't look as nice. If you think it being all uppercase is fine, then I'm happy to change this to what you say, but just thought I would provide my reasoning here.

tobspr: Ah, makes sense. Hm I'm not too sure how much I like the formatting

tobspr:
Will add some changes afterwards - thanks!

PR #48, opened 2020-05-30
Mouse keybinds

Dimava:


src/js/states/keybindings.js
```diff
@@ -121,9 +121,24 @@ export class KeybindingsState extends TextualGameState {
...
+
+            this.app.settings.updateKeybindingOverride(id, keyCode);
 
+            this.dialogs.closeDialog(dialog);
+            this.updateKeybindings();
+        };
+
+        dialog.inputReciever.backButton.add(() => {});
         this.dialogs.internalShowDialog(dialog);
+        dialog.element.onmousedown = clickListener;
```
Dimava, line 140: This one is hacky, but Dialog has no global click listener

src/js/states/keybindings.js
```diff
@@ -121,9 +121,24 @@ export class KeybindingsState extends TextualGameState {
             this.updateKeybindings();
         });
 
-        dialog.inputReciever.backButton.add(() => {});
+        const clickListener = event => {
+            console.log(event);
```
tobspr, line 125: I think you missed a console.log there

src/js/game/camera.js
```diff
@@ -484,6 +485,7 @@ export class Camera extends BasicSerializableObject {
         if (!this.checkPreventDoubleMouse()) {
             return;
         }
+        this.root.app.inputMgr.handleKeyup(event);
```
tobspr, line 488: Whats the reason you are manually calling keyup here?

src/js/states/keybindings.js
tobspr: Could you use addEventListener then?

src/js/game/camera.js
Dimava: Because I haven't found how to bind it correctly that time
Fixed.

tobspr:
Thanks a lot!

PR #58, opened 2020-06-01
fix belt cloning bug

Dimava:
belt_base is the only object with `instantEject: true`, which is not correctly cloned


tobspr:
Nice catch, thank you!

PR #69, opened 2020-06-05
Mod Support

jmeggitt:
This pull request adds the minimum required code to begin creating and testing mods.

Please keep in mind:
- The framework to support downloading mods while in production does not exist yet and will likely be changed in the future
- There are some missing `base-en.yaml` translations that still need to be added
- At the current point, it will not save all mod data and may corrupt savegames if used on an exiting world
- The test mod I used can be found at https://github.com/jmeggitt/shapez.io-modding (More complete information can be found at https://github.com/tobspr/yorg.io-3-modding-docs)
- This pull request is not finished and everything is subject to change!


tobspr:
Is this still being worked on?

PR #72, opened 2020-06-06
Goal amounts on existing pins change when you upgrade

hexagonhexagon:
As is in the title. It has been repeatedly pointed out, so I decided to fix it by adding a new function to `HUDPinnedShapes` and intercepting the existing `upgradePurchased` signal. Feel free to make changes wherever they are necessary.

Note: When you upgrade from Tier V -> Tier VI, the shapes that are removed from the goal are not automatically unpinned by this. They are simply left up as-is, and can be manually clicked away.

### DIFF
--- a/src/js/game/hud/parts/pinned_shapes.js
+++ b/src/js/game/hud/parts/pinned_shapes.js
@@ -89,6 +89,10 @@ export class HUDPinnedShapes extends BaseHUDPart {
         }
     }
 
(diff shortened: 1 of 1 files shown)


hexagonhexagon:
Haven't taken a serious look at this in a bit. What does that mean for the future of this pull request? Will you be refactoring the pinned shapes yourself?

PR #74, opened 2020-06-06
Make selection relative to world

Dimava:
Currently selection is relative to camera, and moves when camera is moved by WASD
This commit fixes that.


src/js/game/hud/parts/mass_selector.js
```diff
@@ -41,6 +41,7 @@ export class HUDMassSelector extends BaseHUDPart {
         this.deletionMarker = Loader.getSprite("sprites/misc/deletion_marker.png");
 
         this.currentSelectionStart = null;
```
tobspr, line 43: Could you change it so it only uses `currentSelectionStartWorld`? No need for two variables

Dimava: fixed

tobspr:
LGTM, thanks a lot!

PR #75, opened 2020-06-07
Add the js step in build.dev

fistons:
While trying to build shapez.io locally to be served by nginx, I noticed a missing js step in the `buid.dev` step

### DIFF
--- a/gulp/gulpfile.js
+++ b/gulp/gulpfile.js
@@ -233,6 +233,7 @@ gulp.task("build.dev", cb => {
         "imgres.copyImageResources",
         "imgres.copyNonImageResources",
         "translations.fullBuild",
+        "js.dev",
         "css.dev",
         "html.dev"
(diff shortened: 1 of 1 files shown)


tobspr:
Well, actually this causes the compilation to run twice when running the serve task. The build.dev step is not supposed to run alone - why do you serve it with nginx? Theres also a webserver command

fistons:
The goal was to be able to build a Docker image with shapez.io in it. When I build it with the embed webserver, the image is ~2.7Go, which is pretty huge I think, so I wanted to build shapez.io for nginx (with an image size of ~35Mo)

I also tries `yarn gulp build.prod` but without success

tobspr:
What didn't work with `build.prod`? It's supposed to work

tobspr:
I'll close this for now, please reopen if its still relevant

PR #85, opened 2020-06-10
Added continue button

jaysc:
I've added a continue button to load the last updated game.

image

- Button will appear only when a save game exists.
- Button will be removed with no save game exists.
- Button CSS is the same as the play button.
- Logic is to load the save file with the newest `lastUpdate` value.

## Reason for change

I kept on pressing "play" expecting to continue where I left of.

## Follow up questions

- Please let me know what CSS to apply for the button itself.
- Are other translation needed.

### DIFF
--- a/src/css/states/main_menu.scss
+++ b/src/css/states/main_menu.scss
@@ -244,6 +244,22 @@
             }
         }
 
+        .continueButton {
+            @include SuperHeading;
+            @include S(min-width, 130px);
+            @include S(padding, 15px, 20px);
+            @include S(margin-top, 15px);
+            letter-spacing: 0.3em !important;
+
+            font-weight: bold;
+            color: #fff;
+            background-color: $colorGreenBright;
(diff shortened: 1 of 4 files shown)


tobspr:
Hey, cool idea!

What if the play button would automatically get a continue button if you have a savegame, and there is a small "New game" button right to the import button?

jaysc:
Upcoming commit will look like this:

image
image

## Changes involved

- I had to create a new `renderMainMenu` function since now it's dynamically adding all the buttons depending on whether there is a save file or not.
- Main menu will re-render when you delete the last save file.
- Added a new property to check if there are save files.
- Added some more helper methods to help with adding elements.
  - `makeButtonBefore` and `makeDivBefore` are not being used. Not sure if you want to keep them for future usage.

tobspr:
Awesome! I'll have to take a closer look. Will take some time but I'll def. get back to this!

tobspr:
Tested, works - thank you!

PR #91, opened 2020-06-10
Add Portugal Portuguese translation

nukuuu:
I've proposed a Portugal Portuguese translation (it's similar but different from Brazilian Portuguese). I'm a native Portuguese speaker but there may be some typos so feel free to edit as you see fit or refuse my contribution. Also keep in mind that this is my first pull request ever.

I believe I have translated everything in this pull but a final proof read is required (I will do it soon).

Pequeno


tobspr:
Awesome thanks  a lot!

PR #92, opened 2020-06-11
Added Czech translations

greeny:
Native Czech speaker. Translations should be at least decent, I assume there will be some fixes needed tho.


tobspr:
Awesome, thanks a lot!

PR #93, opened 2020-06-11
Updated python script.

theisen1337:
I added necessary changes to allow the python script to run in python3, while being backwards compatible with 2.7.


tobspr:
Nice, thanks!

PR #97, opened 2020-06-11
Polished the Polish Translation

mini-bomba:
It appears that we have 2 polish translators. I guess there's now someone that can check your translation for typos and all the other things.
That's exactly what I did - Fixed typos, changed some single words & other stuff. Should be ready to merge.

Also, all subkeys of `global.time` require some extra logic to properly translate. That will be done later.


tobspr:
Nice, thanks!

PR #98, opened 2020-06-11
Change of "Spanish (Latin America)" to just "Spanish"

Al3xFreeman:
I made the Spanish translation and as I used a fairly neutral spanish I think it'd be better to just state that the game is in spanish and not make a separation between both, as it can create a bit of confusion.


tobspr:
Okay, I'll change it later in the UX

PR #102, opened 2020-06-11
Polish the Polish translation even more

mini-bomba:
Summary of changes:

- [x] Added missing line breaks
- [x] Minor translation changes
- [x] Added & translated new translation key: `buildings.hub`


tobspr:
Thanks!

PR #106, opened 2020-06-11
fixed minor inconsistency in painter description

CerxMe:
input sides incorrectly described for the default painter.


tobspr:
Ah, good catch!

PR #108, opened 2020-06-11
Prevent invalid connection if existing underground tunnel entrance exists

jaysc:
This Fixes the issue where it allows a valid underground connection even though an entryway already blocks its path.

## Before

image

## After

image

works as usual when nothing blocks its way

image

## Notes

Original issue found on trello board: https://trello.com/c/4Gp1HC6J


tobspr:
Ah nice, I was actually planning to do that today but I guess this saved me some work - Thank you!

PR #109, opened 2020-06-11
More Russian translation

Killgaru:


Killgaru:
In section "for formatting past time dates"(line 88) not fully correct translate because of some differences between languages. 
In English:
1 		secon**d**
2-60 	second**s**
In Russian:
1 		секунд**у**
2-4 		секунд**ы**
5-20 	секун**д**
21 		секунд**у**
22-24 	секунд**ы**
26-30 	секун**д**
31 		секунд**у**
32-34 	секунд**ы**
35-40 	секун**д**
41 		секунд**у**
42-44 	секунд**ы**
45-50 	секун**д**
51 		секунд**у**
52-54 	секунд**ы**
55-60 	секун**д**
I can combine two ending (what I already do) but I have three.
And it same works with minutes, hours and days.
So, I think what need some algorithm for this, which depending on the time choose the correct ending or word.

tobspr:
oof. Yeah can we maybe find a pragmatic solution?

PR #111, opened 2020-06-11
base-nl.yaml

Ryoncai:
I translated just about everything! Let me know if there is any errors and I'll adjust the text.


tobspr:
Thank you!

PR #113, opened 2020-06-11
Add cut, paste options for blueprints

hexagonhexagon:
Add the ability to cut and paste rather than just copy and paste, as well as add a button to paste the last blueprint you copied or cut. By default, cut is "X" and paste is "V".

Note: The last blueprint you placed isn't preserved when you save/exit or go to the settings menu since there isn't a standard way to serialize blueprints yet.


tobspr:
Very nice! Thank you!

PR #117, opened 2020-06-12
I finished Translating

Jaeyoon07031:
So, I finished translating and finished proofreading it. I will be regularly reading it to find any typos


tobspr:
Awesome, thank you!

PR #124, opened 2020-06-12
Made debug overlay able to be translated

mini-bomba:
The F4 debug overlay (Tickrate, FPS & Tick duration) can now be translated.

I know that this doesn't need to be translated... so there is a comment in base-en.yaml stating this.
Only added English translations for this. At least this makes it a little less hard-coded.


tobspr:
I appreciate the work, but this is really a minor thing - And it adds additional work for translators, who probably will still translate it.

PR #126, opened 2020-06-12
Free Texture packing workflow.

theisen1337:
This allows for an alternate workflow for texture and loading textures into the atlas allowing for more users to create more content for the game.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ This project is based on ES5. Some ES2015 features are used but most of them are
 
 For most assets I use Adobe Photoshop, you can find them in `assets/`.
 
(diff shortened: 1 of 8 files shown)


theisen1337:
This will allow great advances in user create content for the game.

tobspr:
I appreciate the work - But actually I would prefer a more battle tested solution. @dengr1065 Wanted to look into GDX Texture Packager afaik. I feel like having one definite workflow would be better

theisen1337:
Well I really wish you would have send something before I spent all that time on it. perfecting all the sprites settings with the text packer, only took me like 14 hours. It doesn't replace any of the textures by default. It needs to be manually ran than the project needs to be rebuilt..

tobspr:
Closing this for now

PR #130, opened 2020-06-12
Translate newly added translation keys to Polish

mini-bomba:
Translated those newly added translation strings to Polish. Also fixed a minor mistake & re-translated one string that somehow got lost.


tobspr:
Thank you!

PR #131, opened 2020-06-12
added shift modifier for faster pan

jaysc:
Added ability to move faster by holding shift.

Based on the existing movement settings code, this will check whether you are holding the `shift` key, then go to the next level of movement speed. If you are already set to the maximum, it won't go faster.

Tested on all speeds.

### DIFF
--- a/src/js/profile/application_settings.js
+++ b/src/js/profile/application_settings.js
@@ -302,14 +302,11 @@ export class ApplicationSettings extends ReadWriteProxy {
         return 1;
     }
 
-    /**
-     * @param {boolean} moveFaster
-     */
-    getMovementSpeed(moveFaster = false) {
+    getMovementSpeed() {
         const id = this.getAllSettings().movementSpeed;
         for (let i = 0; i < movementSpeeds.length; ++i) {
             if (movementSpeeds[i].id === id) {
-                return movementSpeeds[i].multiplier * (moveFaster ? 2 : 1);
+                return movementSpeeds[i].multiplier;
             }
         }
         logger.error("Unknown movement speed id:", id);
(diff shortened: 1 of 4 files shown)


tobspr:
Looks good! Could you add a translation in base-en for it?

jaysc:
- Renamed enum from `fastPan` to `mapMoveFaster` for consistency.
- Added translation string

image

src/js/profile/application_settings.js
```diff
@@ -302,11 +302,14 @@ export class ApplicationSettings extends ReadWriteProxy {
...
-    getMovementSpeed() {
+    /**
+     * @param {boolean} moveFaster
+     */
+    getMovementSpeed(moveFaster = false) {
         const id = this.getAllSettings().movementSpeed;
         for (let i = 0; i < movementSpeeds.length; ++i) {
             if (movementSpeeds[i].id === id) {
-                return movementSpeeds[i].multiplier;
+                return movementSpeeds[moveFaster && i + 1 < movementSpeeds.length ? i + 1 : i].multiplier;
```
tobspr, line 312: Not too sure about this, i think moveFaster should just be * 2 or so

jaysc: I was thinking that as a simpler way to handle it. The speed steps are actually doubled anyway. I'll just do 2 * in the next commit.

src/js/profile/application_settings.js
```diff
@@ -302,11 +302,14 @@ export class ApplicationSettings extends ReadWriteProxy {
...
-    getMovementSpeed() {
+    /**
+     * @param {boolean} moveFaster
+     */
+    getMovementSpeed(moveFaster = false) {
         const id = this.getAllSettings().movementSpeed;
         for (let i = 0; i < movementSpeeds.length; ++i) {
             if (movementSpeeds[i].id === id) {
-                return movementSpeeds[i].multiplier;
+                return movementSpeeds[i].multiplier * (moveFaster ? 2 : 1);
```
tobspr, line 312: Last annotation (Sorry for being so picky again!): I think it doesn't really belong in application_settings - I think it would fit much better in the camera class

jaysc: removed from application_settings and merged it with the camera.js settings

tobspr:
Awesome, thank you! Will be in 1.1.13

PR #132, opened 2020-06-12
Made About this game translatable

mini-bomba:
Made the `About this game` section body translatable. It looks like it wouldn't change, so I added a translation key to be able to translate this to all supported languages.

Quick summary:
- [x] Added new translation key: `about.body` - For translating the `About this game` section's body
- [x] Translated `about.body` to Polish
- [x] Tested if this works - it works.


tobspr:
Thank you! Kinda makes sense, I admit I just didn't have time

PR #133, opened 2020-06-12
Polishing The Polish Translation™ #6

mini-bomba:
Translated some words that were missed before & re-translated keys that somehow got reset.
And some more minor changes as always.


tobspr:
Thank you!

PR #134, opened 2020-06-12
Fix keybindings button label text not changing with language

mini-bomba:
Before this PR the keybindings button (in settings) had the same label text, regardless of selected language. This PR fixes that. It's label text is now pulled from translations in the same way, as the `About this game` button does (which is next to it and didn't have this bug). This is a simple, 1-line fix.


tobspr:
Nice, thanks!

PR #135, opened 2020-06-12
Update base-nl.yaml

Ryoncai:
Made some minor tweaks, translated a few short sections I accidentally skipped over earlier and translated the new stuff.

I'm not certain if line 667 should be translated (to 'sneltoetsen'). I thought not, but I noticed that the blue tab ingame under options still says keybindings, and I'm not sure how else to change that one.


tobspr:
Thank you!

PR #138, opened 2020-06-13
Redo stacking algorithm

hexagonhexagon:
Many people have mentioned that stacking for the logo shape feels unintuitive, so I have redone the stacking algorithm so that instead of merging the top shape down layer by layer, it preserves the top shape whole and stacks that down as far as it can instead. The results are displayed below. Feel free to request changes/comments as necessary, or just reject this entirely if you don't like it.

image

### DIFF
--- a/src/js/game/shape_definition.js
+++ b/src/js/game/shape_definition.js
@@ -536,7 +536,8 @@ export class ShapeDefinition extends BasicSerializableObject {
             gapsBetweenShapes.push(topShapeLowestLayerByQuad[quad] - bottomShapeHighestLayerByQuad[quad]);
         }
         const smallestGapBetweenShapes = Math.min(...gapsBetweenShapes);
-        const layerToMergeAt = 1 - smallestGapBetweenShapes;
+        // Can't merge at a layer lower than 0
+        const layerToMergeAt = Math.max(1 - smallestGapBetweenShapes, 0);
 
         const mergedLayers = this.internalCloneLayers();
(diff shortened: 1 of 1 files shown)


tobspr:
This has been on my todo list for ages - but it will also take some time to review the whole logic :)

TheBosZ:
Will this also fix stacking an already stacked item?
image

(I can file an issue if it's unrelated)

hexagonhexagon:
That's happening because the maximum number of layers a shape can have is 4, and when you get more than that by stacking, the game chops off all layers above the 4th. I have decided to preserve that behavior in this PR, but if you want it to not do that feel free to make an issue.

dengr1065:
it should be left as-is, recursive shape stacking may break the game otherwise

TheBosZ:
> That's happening because the maximum number of layers a shape can have is 4, and when you get more than that by stacking, the game chops off all layers above the 4th. I have decided to preserve that behavior in this PR, but if you want it to not do that feel free to make an issue.

You're totally right! I was trying to get this and I was mistaken: 
image

tobspr:
I really should merge this. But I don't want to break existing factories which is why I will wait until the wires update

hexagonhexagon:
Understandable. Take your time.

hexagonhexagon:
The latest commit solves a possible crash if #352 is merged along with this request, in which case it may happen that `layerToMergeAt` can be less than 0, which is normally not possible under current cutting rules, but found out it could happen when playing with a change similar to this.

Dimava:
I suggest to already merge this PR into branch with wires

tobspr:
I didn't test it. yolo

PR #140, opened 2020-06-13
Dark theme for changelog and completed upgrades

dengr1065:
really quick fix just 3 sloc

### DIFF
--- a/src/css/ingame_hud/shop.scss
+++ b/src/css/ingame_hud/shop.scss
@@ -241,6 +241,10 @@
 
                             &.complete {
                                 background-color: $colorGreenBright;
+                                
+                                @include DarkThemeOverride {
+                                    background-color: $colorGreenBright;
+                                }
                             }
                         }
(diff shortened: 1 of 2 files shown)


src/css/ingame_hud/shop.scss
```diff
@@ -241,6 +241,10 @@
 
                             &.complete {
                                 background-color: $colorGreenBright;
+                                
+                                @include DarkThemeOverride {
+                                    background-color: $colorGreenBright;
+                                }
```
tobspr, line 247: What is this doing here?

dengr1065: DarkThemeOverride for parent (`progressBar` or something like that) for some reason takes over `.complete` background color, even though it should not; so applying color here as well fixes it.

dengr1065:
marked as resolved because no feedback

tobspr:
Thank you!

PR #151, opened 2020-06-14
A humble idea for color blind information. Not ready to merge. [PoC]

CaptainDapper:
Had this thought while falling asleep the other night. I've never worked with nodejs or with the HTML canvas so I thought I could learn a thing or two. I did! Hopefully this is a useful idea, although it seems rather difficult to see, especially while things are in motion.

### DIFF
--- a/src/js/game/shape_definition.js
+++ b/src/js/game/shape_definition.js
@@ -346,6 +346,82 @@ export class ShapeDefinition extends BasicSerializableObject {
                 context.fill();
                 context.stroke();
 
+                let draw_vert = false;
+                let draw_pos = false;
+                let draw_neg = false;
+
+                switch (color) {
+                    case "red":
+                        draw_vert = true;
+                        break;
+                    case "green":
+                        draw_pos = true;
+                        break;
+                    case "blue":
+                        draw_neg = true;
+                        break;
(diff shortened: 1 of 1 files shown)


CaptainDapper:
image

tobspr:
Hey, this is a cool idea! However, i see the following issue - What if you are stacking items?

CaptainDapper:
That's a fair point! I hadn't gotten that far, I guess :P Besides the "Hard to see while moving" issue, I also found myself wishing I could zoom in farther (maybe to 5 instead of 3?) so it was easier to see the hatching, but I think the same thing would apply for your #155 idea.

No matter what, this is a really really complicated issue. To get something going, it might be easy enough (from a coding standpoint) just to change the colors (in a menu option) to something more suitable for the most common types of colorblindness. You can simulate them with this Chrome Extension: [Colorblindly](https://chrome.google.com/webstore/detail/colorblindly/floniaahmccleoclneebhhmnjgdfijgg/related) and there is more information about the statistics of colorblindness [here](https://iristech.co/statistics/). That's probably harder than it sounds to tune just right though...

tobspr:
This has been solved as of 1.1.17

PR #153, opened 2020-06-14
Add dye incinerator building

hexagonhexagon:
Add a building to convert any dye to uncolored dye, primarily for use in the quad painter, but can also be used for freeplay things. I noticed a bunch of people asking for the quad painter to be better, so this is my solution. This has a hub unlock attached to it; it unlocks right after the quad painter for 150k of the 4-layer pinwheel shape.
image


tobspr:
I feel like its overkill to have this as seperate building

tobspr:
Closing this for now, since I have no plans to merge it in the near future

PR #155, opened 2020-06-14
[PoC] Testing another approach for a colorblind modus

attrib:
After seeing the PoC  (https://github.com/tobspr/shapez.io/pull/151) from @CaptainDapper I had another idea.

This will also not solve for stacked items issue mentioned by @tobspr in the comments of #151, but maybe with more ideas a solution could be found.

My idea here is that for the color items you have the 3 circles and each circle represents one color (green, red, blue). Only colors used for the mix are filled, the others are not filled.

On the items I just add the circles again, to identify which mixture is needed.

Regarding the color changes, these help me as the contrast is higher and I'm able to see the difference a bit better. But this is my personal experience, for others this result could be different!

### DIFF
--- a/src/js/game/colors.js
+++ b/src/js/game/colors.js
@@ -35,14 +35,14 @@ for (const key in enumColorToShortcode) {
 /** @enum {string} */
 export const enumColorsToHexCode = {
     [enumColors.red]: "#ff666a",
-    [enumColors.green]: "#78ff66",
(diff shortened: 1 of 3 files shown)


attrib:
Color Items
shapz-color

Item with Color
shapz-color-item

tobspr:
Closing because 1.1.17 will have a solution for this

PR #156, opened 2020-06-14
Add keybind to enable/disable vignette effect.

hexagonhexagon:
By default, the button is F5 (displays as F4 ingame due to a bug). Useful for taking building screenshots. I also personally don't like vignette effects, but here everything is way too bright when you disable it.


tobspr:
I think this should be a setting instead

hexagonhexagon:
Closing as a setting has been added for this.

PR #159, opened 2020-06-14
Fix camera inertia problem with mouse movement

hexagonhexagon:
Currently when moving the camera with the mouse, there is an annoying feature that if you hold the camera still and then let go, there is some stored velocity that gets applied after you let go instead of it just staying still where you let go. I fix this behavior by adding some code to `internalUpdatePanning()` that checks if `lastMovingPosition` has been constant for the last 2 ticks: if it has, then zero out `touchPostMoveVelocity`. 

Note: I have not tested this on touch devices, and that should probably be done before pulling this.

### DIFF
--- a/src/js/game/camera.js
+++ b/src/js/game/camera.js
@@ -28,6 +28,7 @@ const velocitySmoothing = 0.5;
 const velocityFade = 0.98;
 const velocityStrength = 0.4;
 const velocityMax = 20;
+const ticksBeforeErasingVelocity = 2;
 
 /**
  * @enum {string}
@@ -58,6 +59,8 @@ export class Camera extends BasicSerializableObject {
         // Input handling
         this.currentlyMoving = false;
         this.lastMovingPosition = null;
+        this.lastMovingPositionLastTick = null;
+        this.numTicksStandingStill = null;
         this.cameraUpdateTimeBucket = 0.0;
         this.didMoveSinceTouchStart = false;
         this.currentlyPinching = false;
@@ -667,6 +670,8 @@ export class Camera extends BasicSerializableObject {
         this.touchPostMoveVelocity = new Vector(0, 0);
         this.currentlyMoving = true;
         this.lastMovingPosition = pos;
+        this.lastMovingPositionLastTick = null;
+        this.numTicksStandingStill = 0;
         this.didMoveSinceTouchStart = false;
     }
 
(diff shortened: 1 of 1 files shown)


tobspr:
Uh can't we find a more elegant solution? :P

hexagonhexagon:
Would you mind clarifying what feels so "hacky" about this? I'm trying to take a serious look at this to see how I could do it differently, but I'm not seeing what I could improve in order to do this.

src/js/game/camera.js
```diff
@@ -712,6 +717,8 @@ export class Camera extends BasicSerializableObject {
             this.currentlyMoving = false;
             this.currentlyPinching = false;
             this.lastMovingPosition = null;
+            this.lastMovingPositionLastTick = null;
+            this.numTicksStandingStill = null;
```
tobspr, line 721: I think we should make it consistent here and set it to `0` instead?

src/js/game/camera.js
```diff
@@ -809,6 +816,20 @@ export class Camera extends BasicSerializableObject {
 
         this.touchPostMoveVelocity = this.touchPostMoveVelocity.multiplyScalar(velocityFade);
 
+        // Check if the camera is being dragged but standing still: if not, zero out `touchPostMoveVelocity`.
+        if (this.currentlyMoving && this.desiredCenter === null) {
+            if (this.lastMovingPositionLastTick === this.lastMovingPosition) {
+                this.numTicksStandingStill += 1;
```
tobspr, line 822: could use `++` here

src/js/game/camera.js
```diff
@@ -809,6 +816,20 @@ export class Camera extends BasicSerializableObject {
 
         this.touchPostMoveVelocity = this.touchPostMoveVelocity.multiplyScalar(velocityFade);
 
+        // Check if the camera is being dragged but standing still: if not, zero out `touchPostMoveVelocity`.
+        if (this.currentlyMoving && this.desiredCenter === null) {
+            if (this.lastMovingPositionLastTick === this.lastMovingPosition) {
```
tobspr, line 821: Better use `.equalsEpsilon` here

src/js/game/camera.js
```diff
@@ -809,6 +816,20 @@ export class Camera extends BasicSerializableObject {
...
+                this.numTicksStandingStill += 1;
+            } else {
+                this.numTicksStandingStill = 0;
+            }
+            this.lastMovingPositionLastTick = this.lastMovingPosition;
```
tobspr, line 826: Could you use `.copy()` here? Otherwise you will share the same vector reference

tobspr:
I guess its fine, made some annotations tho

hexagonhexagon:
The changes you requested have been made.

tobspr:
Nice, thank you!

PR #160, opened 2020-06-14
Czech translation update

MrMates:
I checked for some new strings to translate, also translated those that were still in English.

I also gave a shot to translating the "About Game" part as well... **I tried to play with it** – to save the original point of "you (the dev)" references – by talking about the dev in 3rd person and keeping it warm and cozy for native Czech peeps :)
That's obviously just an idea, keep it in English if you want to leave it personal


tobspr:
Awesome, thank you! :) About section is fine :)

PR #162, opened 2020-06-15
Fix bug where belts in blueprints don't orient correctly.

hexagonhexagon:
The bug: copying the left hand side results in the right hand side when pasted.
image

The problem: This is happening because as part of `EntityManager.registerEntity`, the `entityAdded` signal is raised immediately, which then is received by `BeltSystem.updateSurroundingBeltPlacement`, which is a problem since we are reorienting the belt before all the belts have been placed down, resulting in the belt changing from the correct direction to the wrong direction. 

The fix: We simply don't register belt entities until all entities from the blueprint have been placed down, ensuring that `BeltSystem.updateSurroundingBeltPlacement` won't change the direction of the belt since all the belts are actually down.

### DIFF
--- a/src/js/game/hud/parts/blueprint.js
+++ b/src/js/game/hud/parts/blueprint.js
@@ -203,10 +203,10 @@ export class Blueprint {
                                     "Can not delete entity for blueprint"
                                 );
                                 if (!root.logic.tryDeleteBuilding(contents)) {
-                                    logger.error(
+                                    assertAlways(
+                                        false,
                                         "Building has replaceable component but is also unremovable in blueprint"
                                     );
-                                    return false;
                                 }
                             }
                         }
--- a/src/js/game/hud/parts/blueprint.js
+++ b/src/js/game/hud/parts/blueprint.js
@@ -176,6 +176,7 @@ export class Blueprint {
     tryPlace(root, tile) {
         return root.logic.performBulkOperation(() => {
             let anyPlaced = false;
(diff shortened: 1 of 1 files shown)


src/js/game/hud/parts/blueprint.js
```diff
@@ -215,10 +216,22 @@ export class Blueprint {
...
+                        root.entityMgr.registerEntity(clone);
+                    } else {
+                        beltsToRegisterLater.push(clone);
+                    }
                     anyPlaced = true;
                 }
             }
+
+            for (let i = 0; i < beltsToRegisterLater.length; i++) {
+                root.entityMgr.registerEntity(beltsToRegisterLater[i]);
```
tobspr, line 233: This looks like a solution, however:
In line 209 a return is possible - This means the entities would have been placed on the map but not registered (Which will lead to a crash). Either we make this a hard assertion (`assertAlways`) so this is a crash anyways, or we make sure that we register the already placed entities there

hexagonhexagon: I think making line 209 an `assertAlways` makes the most sense here, since I can't think of a scenario where you would want a building would be both replaceable and unremovable.

tobspr: Good idea

tobspr:
Would be nice to have this for 1.1.17

tobspr:
Awesome, thanks!

PR #168, opened 2020-06-15
Polishing The Polish Translation™ #7

mini-bomba:
This PR does some minor changes to the Polish translation (as always) and adds some missed/reset translations (also as always).


tobspr:
Nice, thank you! This must be really polished now :D

PR #176, opened 2020-06-16
Update base-nl.yaml

Milomilli:
I changed some words where I see it is been trough google translate, I also changed "lopende banden" into "banden" cause it overlaps with other words
I hope you have something with it


tobspr:
Alright, thank you!

PR #178, opened 2020-06-16
Update Japanese translation

kebhr:
- Update Japanese translation
- Recover lost translations (due to mismatching placeholders)


tobspr:
Thank you!

PR #186, opened 2020-06-16
Fixed standalone only setting hint not being translated

mini-bomba:
This PR fixes that "Not available in demo" text always being in english, even if english wasn't the selected language.
This does not add any new translation keys, as one was already present.


tobspr:
I don't get it - you just moved the code?

tobspr:
Ahh I get it. It was not translated initially. My bad

tobspr:
Thank you!

PR #187, opened 2020-06-16
German suffix for large numbers corrected

DecaturRamsey:
Changed the suffix for large numbers to reflect the German naming convention.
million -> "Millionen" (Mio)
billion-> "Milliarden" (Mrd)
trillion -> "Billionen" (Bio)

### DIFF
--- a/translations/base-de.yaml
+++ b/translations/base-de.yaml
@@ -76,9 +76,9 @@ global:
     # The suffix for large numbers, e.g. 1.3k, 400.2M, etc.
     suffix:
         thousands: T
-        millions: M
-        billions: B
-        trillions: tr
+        millions: Mio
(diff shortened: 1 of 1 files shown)


tobspr:
Das ist zu lang in der UI

DerFeldspatz:
I would suggest changing it to metric prefixes (k,M,G,T,P) as it is used on german Wikipedia

tobspr:
Good idea

tobspr:
TBH The user will see "Million" at max. Nobody's going to have a billion of one shape

dengr1065:
except those who have powerful machines and generate 50k of blueprint shapes

tobspr:
Thank you!

PR #192, opened 2020-06-16
Fix crash when deleting buildings.

grickit:
Can't check getHasDirectionLockAvailable() on a building you are currently deleting because it doesn't exist.


ropiku:
Can reproduce and confirm the fix works for me

tobspr:
Nice thanks! Actually this will be obsolete since I'll rewrite that code, but thanks anyways :P

PR #193, opened 2020-06-16
[Minor Opt] Make miners cache mined item

Phlosioneer:
Saves two chunk+tile lookups per update, and one chunk+tile lookup per draw.

### DIFF
--- a/src/js/game/systems/miner.js
+++ b/src/js/game/systems/miner.js
@@ -24,10 +24,8 @@ export class MinerSystem extends GameSystemWithFilter {
             // Check if miner is above an actual tile
 
             const minerComp = entity.components.Miner;
-            const staticComp = entity.components.StaticMapEntity;
 
-            const tileBelow = this.root.map.getLowerLayerContentXY(staticComp.origin.x, staticComp.origin.y);
-            if (!tileBelow) {
+            if (!minerComp.minedItem) {
                 continue;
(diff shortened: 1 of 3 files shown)


tobspr:
This will break existing savegames, it would be required to write a migration then (Check out how 1003 -> 1004 works for example. simply search for `1004` in the code). Looks good otherwise!

tobspr:
I would recommend not to serialize/save this - Instead make the miner system compute it whenever an entity gets added (root.signals.entityAdded)

Phlosioneer:
Yeah, I learned I could avoid serialization on the later PR's. It makes more sense here.

src/js/game/systems/miner.js
```diff
@@ -114,18 +109,17 @@ export class MinerSystem extends GameSystemWithFilter {
...
+                        continue;
+                    }
 
-                    const lowerLayerItem = this.root.map.getLowerLayerContentXY(
-                        staticComp.origin.x,
-                        staticComp.origin.y
-                    );
-
-                    if (lowerLayerItem) {
+                    if (minerComp.cachedMinedItem) {
```
tobspr, line 120: This if is unnecessary, since you continue above. I can remove it myself tho

tobspr:
Thank you! Will be in 1.1.16

PR #196, opened 2020-06-16
Updated Norwegian Translastion to newest version

Baxxster:
Updated Norwegian Translastion to newest version


tobspr:
Nice, thank you!

PR #198, opened 2020-06-16
Optimize belt cache

Phlosioneer:
Removed the belt cache array. Follow-up belts are cached in the belt's BeltComponent instead. This change also removes the recursive follow-up search, which could cause a stack overflow for an extremely long belt chain.

Saves one object allocation per belt per change, two very large array allocations per change, many function calls, and belts are only visited exactly once per change.


tobspr:
Very nice as well! Testing it soon

tobspr:
I target this to be merged for the next major performance update

tobspr:
I'll make some adjustments but the base is fine - thank you!

PR #205, opened 2020-06-17
Updated french translation

PascalGrosse:


tobspr:
There is a conflict because somebody else already translated parts of it, could you have a look? :)

tobspr:
Thanks!

PR #206, opened 2020-06-17
Update pt-pt translation to the latest version

nukuuu:
Updated pt-pt translation to the latest version and corrected two translations

This updated has been made with discord user Gumball73#0121

Pequeno


tobspr:
Thank you!

PR #212, opened 2020-06-17
fix for screenshots larger than 16,384 x 16,384

xSparfuchs:
dynamically adjusts chunkSizePixels to fit all chunks in 16,384 x 16,384 pixels

### DIFF
--- a/src/js/game/hud/parts/screenshot_exporter.js
+++ b/src/js/game/hud/parts/screenshot_exporter.js
@@ -58,9 +58,9 @@ export class HUDScreenshotExporter extends BaseHUDPart {
 
         let chunkSizePixels = 128;
         const maxDimensions = Math.max(dimensions.x, dimensions.y);
-        
+
         if (maxDimensions > 128) {
-                chunkSizePixels = Math.floor(128 * (128 / Math.max(dimensions.x, dimensions.y)));
+                chunkSizePixels = Math.floor(128 * (128 / maxDimensions));
         }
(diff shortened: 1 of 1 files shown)


tobspr:
I think you mixed tabs an spaces :)

xSparfuchs:
looks like i did an oopsie there, should be fine now

src/js/game/hud/parts/screenshot_exporter.js
```diff
@@ -56,7 +56,17 @@ export class HUDScreenshotExporter extends BaseHUDPart {
         const dimensions = maxChunk.sub(minChunk);
         logger.log("Dimensions:", dimensions);
 
-        const chunkSizePixels = 128;
+        let chunkSizePixels = 128;
+
+        if (dimensions.x + dimensions.y > 256) {
+            if (dimensions.x > dimensions.y) {
+                chunkSizePixels = Math.floor(128 * (128 / dimensions.x));
```
tobspr, line 63: Why not:

`chunkSizePixels = Math.floor(128 * (128 / Math.max(dimensions.x, dimensions.y)));` 

Also I think it'd be simpler to do:

`const maxDimensions = Math.max(dimensions.x, dimensions.y)`;

Then:

`if (maxDimensions > 128)`

And then divide by `maxDimensions`

xSparfuchs: yup, that makes a lot of sense

tobspr:
Thank you!

tobspr:
Will be in 1.1.16

PR #216, opened 2020-06-17
Implemented /kb_enter and /kb_escape on Dialogs

Andifaind:
Added this to default values for "**showInfo**" method
Added this to "**Delete**" & "**Cut**" confirmation dialogs


tobspr:
This is duplicate functionality -> You can already do `ok:good:enter` and `cancel:bad:escape`. Am I missing someting besides of that?

If not, feel free to make a PR which adds that :)

tobspr:
Obsolete as 1.1.17 added this

Issue #189, opened 2020-06-16
Minor bug: Glitch when simultaneously right and left clicking


Phlosioneer:
I had the upgrades screen opened and moused over the X. I then accidentally simultaneously (or nearly so) clicked with the right and left buttons.

The upgrades window close button was unresponsive to any more clicks.

Pressing the F key to toggle the upgrades window worked. Further use of the upgrades window works fine.

tobspr:
Fixed by that pr :)

PR #218, opened 2020-06-18
Clumsy clicker fix

davidburhans:
Possible fix for Issue #189 

Reproduce issue by pressing left-mouse button over UI button. While left-button is depressed, click right-button. Release left-button.  UI button no longer responds to clicks. Console warning `Ignoring double click` is visible when left-button is clicked on UI button.

Solution:
Since right-mouse button sets `ClickDetector.cancelled = true` the mouse-up handler is never able to clear `ClickDetector.clickDownPosition`. So clear `clickDownPosition` when `cancelled` is set to `true`.


tobspr:
Thanks! Will be in 1.1.16

PR #221, opened 2020-06-18
Update base-nl.yaml

Ryoncai:
Added new translations, fixed some small errors, reversed some unnecessary/confusing translations, made the translations of the other person more consistent with the rest of the text.


tobspr:
Thanks!

PR #223, opened 2020-06-18
Map markers' labels are way too short.

### LINKED ISSUE #224: Map markers' labels are way too short
Max length of 14 characters for map markers' labels is too short. Something simple as "Reserved for CuCuCuCu" or "Next teal squares" won't fit.

It's a single player game after all, if the player doesn't mind very long labels spanning half the screen/map, then what's the harm?

#223

Joker-vD:
Max length of 14 characters for map markers' labels is too short. Something simple as "Reserved for CuCuCuCu" or "Next teal squares" won't fit.

Fixes #224


tobspr:
Can't argue against that

tobspr:
Will be in 1.1.16

PR #227, opened 2020-06-18
Rotation remembered by building type

Magos:
While playing the early game I noticed I was often needing to place painters in one direction and then place mixers in the opposite direction. That caused a lot of rotation key presses that kinda felt like busywork. 
Obviously blueprints eventually make this less of a hassle, but I still thought a setting allowing rotation to be set per building type instead of globally would make sense.

### DIFF
--- a/src/js/game/hud/parts/building_placer_logic.js
+++ b/src/js/game/hud/parts/building_placer_logic.js
@@ -128,9 +128,9 @@ export class HUDBuildingPlacerLogic extends BaseHUDPart {
         if (!this.root.app.settings.getAllSettings().rotationByBuilding) {
             return this.currentBaseRotationGeneral;
         }
-        const building = this.currentMetaBuilding.get();
-        if (building != undefined && this.preferredBaseRotations.hasOwnProperty(building.getId())) {
-            return this.preferredBaseRotations[building.getId()];
+        const metaBuilding = this.currentMetaBuilding.get();
+        if (metaBuilding && this.preferredBaseRotations.hasOwnProperty(metaBuilding.getId())) {
+            return this.preferredBaseRotations[metaBuilding.getId()];
         } else {
             return this.currentBaseRotationGeneral;
         }
@@ -138,14 +138,15 @@ export class HUDBuildingPlacerLogic extends BaseHUDPart {
 
     /**
      * Sets the base rotation for the current meta-building.
+     * @param {number} rotation The new rotation/angle.
      */
     set currentBaseRotation(rotation) {
         if (!this.root.app.settings.getAllSettings().rotationByBuilding) {
             this.currentBaseRotationGeneral = rotation;
         } else {
-            const building = this.currentMetaBuilding.get();
-            if (building != undefined) {
-                this.preferredBaseRotations[building.getId()] = rotation;
+            const metaBuilding = this.currentMetaBuilding.get();
+            if (metaBuilding) {
+                this.preferredBaseRotations[metaBuilding.getId()] = rotation;
             } else {
                 this.currentBaseRotationGeneral = rotation;
             }
--- a/src/js/game/hud/parts/building_placer_logic.js
+++ b/src/js/game/hud/parts/building_placer_logic.js
@@ -49,28 +49,9 @@ export class HUDBuildingPlacerLogic extends BaseHUDPart {
 
         /**
          * The current rotation preference for each building.
-         * @type {Object.<string, number>}
+         * @type{Object.<string,number>}
          */
-        this.preferredRotations = {};
-
-        this.getBaseRotation = function () {
-            const rotationByBuilding = this.root.app.settings.getAllSettings().rotationByBuilding;
-            if (!rotationByBuilding) {
-                return this.currentBaseRotationGeneral;
-            }
-            const id = this.currentMetaBuilding.get().getId();
-            return this.preferredRotations[id] || this.currentBaseRotationGeneral;
-        };
-
-        this.setBaseRotation = function (rotation) {
-            const rotationByBuilding = this.root.app.settings.getAllSettings().rotationByBuilding;
-            if (!rotationByBuilding) {
-                this.currentBaseRotationGeneral = rotation;
-            } else {
-                const id = this.currentMetaBuilding.get().getId();
-                this.preferredRotations[id] = rotation;
-            }
-        };
+        this.preferredBaseRotations = {};
 
         /**
          * Whether we are currently dragging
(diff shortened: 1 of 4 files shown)


src/js/game/hud/parts/building_placer_logic.js
```diff
@@ -44,7 +44,32 @@ export class HUDBuildingPlacerLogic extends BaseHUDPart {
...
-        this.currentBaseRotation = 0;
+        this.currentBaseRotationGeneral = 0;
+
+        /**
+         * The current rotation preference for each building.
+         * @type {Object.<string, number>}
+         */
+        this.preferredRotations = {};
+
+        this.getBaseRotation = function () {
```
tobspr, line 55: The game uses es6 classes, could you convert that into es6 class functions? Or am I missing something here?

src/js/game/hud/parts/building_placer_logic.js
```diff
@@ -194,12 +219,12 @@ export class HUDBuildingPlacerLogic extends BaseHUDPart {
         const selectedBuilding = this.currentMetaBuilding.get();
         if (selectedBuilding) {
             if (this.root.keyMapper.getBinding(KEYMAPPINGS.placement.rotateInverseModifier).pressed) {
-                this.currentBaseRotation = (this.currentBaseRotation + 270) % 360;
+                this.setBaseRotation((this.getBaseRotation() + 270) % 360);
```
tobspr, line 222: If you change the code into `get currentBaseRotation() {}` you don't have to touch the remaining code. Can also use setters there

translations/base-en.yaml
```diff
@@ -687,6 +687,11 @@ settings:
             title: Vignette
             description: >-
                 Enables the vignette which darkens the screen corners and makes text easier to read.
+
```
tobspr, line 690: Nice, a setting is always good!

src/js/game/hud/parts/building_placer_logic.js
Magos: Yeah, I wasn't very familiar with this syntax in JS. I made the switch now.

src/js/game/hud/parts/building_placer_logic.js
```diff
@@ -114,6 +120,34 @@ export class HUDBuildingPlacerLogic extends BaseHUDPart {
...
+    /**
+     * Returns the current base rotation for the current meta-building.
+     * @returns {number}
+     */
+    get currentBaseRotation() {
+        const rotationByBuilding = this.root.app.settings.getAllSettings().rotationByBuilding;
+        if (!rotationByBuilding) {
+            return this.currentBaseRotationGeneral;
+        }
+        const id = this.currentMetaBuilding.get().getId();
```
tobspr, line 132: .get() can be null, in which case this would throw an error

src/js/game/hud/parts/building_placer_logic.js
```diff
@@ -114,6 +120,34 @@ export class HUDBuildingPlacerLogic extends BaseHUDPart {
         this.root.camera.upPostHandler.add(this.onMouseUp, this);
     }
 
+    /**
+     * Returns the current base rotation for the current meta-building.
+     * @returns {number}
+     */
+    get currentBaseRotation() {
+        const rotationByBuilding = this.root.app.settings.getAllSettings().rotationByBuilding;
+        if (!rotationByBuilding) {
```
tobspr, line 129: I think it'd be fine to put this into one line without a variable

src/js/game/hud/parts/building_placer_logic.js
```diff
@@ -114,6 +120,34 @@ export class HUDBuildingPlacerLogic extends BaseHUDPart {
...
+            return this.currentBaseRotationGeneral;
+        }
+        const id = this.currentMetaBuilding.get().getId();
+        return this.preferredBaseRotations[id] == null
```
tobspr, line 133: If `this.preferredBaseRotations[id]` is `0`, comparing it with `null` will be truthy.

Instead do if `typeof this.preferredBaseRotations[id] === 'undefined'` or even better: this.preferredBaseRotations.hasOwnProperty(id).

Since you are using a map here though, actually the *best* solution would be simply to use a `new Map()`

src/js/game/hud/parts/building_placer_logic.js
Magos: Moved both instances of this to one-liners.

src/js/game/hud/parts/building_placer_logic.js
Magos: I added a check for == undefined before calling getId(), defaulting to currentBaseRotationGeneral when currentMetaBuilding isn't set.

src/js/game/hud/parts/building_placer_logic.js
Magos: I ended up going for the hasOwnProperty() option.

src/js/game/hud/parts/building_placer_logic.js
```diff
@@ -114,6 +120,38 @@ export class HUDBuildingPlacerLogic extends BaseHUDPart {
...
+    /**
+     * Returns the current base rotation for the current meta-building.
+     * @returns {number}
+     */
+    get currentBaseRotation() {
+        if (!this.root.app.settings.getAllSettings().rotationByBuilding) {
+            return this.currentBaseRotationGeneral;
+        }
+        const building = this.currentMetaBuilding.get();
+        if (building != undefined && this.preferredBaseRotations.hasOwnProperty(building.getId())) {
```
tobspr, line 132: I think it would be consistent to just use `if (building && ...)` since this is how its done in the rest of the codebase.

It'd also be nice if you'd use `metaBuilding` as a variable name for it since thats the convention

src/js/game/hud/parts/building_placer_logic.js
```diff
@@ -114,6 +120,38 @@ export class HUDBuildingPlacerLogic extends BaseHUDPart {
...
+            return this.preferredBaseRotations[building.getId()];
+        } else {
+            return this.currentBaseRotationGeneral;
+        }
+    }
+
+    /**
+     * Sets the base rotation for the current meta-building.
+     */
```
tobspr, line 141: Could you add a jsdoc `@param {number} rotation` here? (Sorry for nitpicking :P)

src/js/game/hud/parts/building_placer_logic.js
```diff
@@ -114,6 +120,38 @@ export class HUDBuildingPlacerLogic extends BaseHUDPart {
...
+    set currentBaseRotation(rotation) {
+        if (!this.root.app.settings.getAllSettings().rotationByBuilding) {
+            this.currentBaseRotationGeneral = rotation;
+        } else {
+            const building = this.currentMetaBuilding.get();
+            if (building != undefined) {
```
tobspr, line 147: See above :)

src/js/game/hud/parts/building_placer_logic.js
Magos: Implemented both of these in both functions.

src/js/game/hud/parts/building_placer_logic.js
Magos: Added. No worries, consistent code style is quite helpful for new contributors trying to find their way around.

tobspr:
Thank you!

tobspr:
Will be in 1.1.18

Issue #230, opened 2020-06-18
Smart tunnels removing wrong belts


Andifaind:
Looks like smart tunnels are removing belts between the begin and the end that are facing in the same direction as the tunnel, but it's not always appropriate.

Like in this case:
![tunnel-error](https://user-images.githubusercontent.com/67031075/85082060-2b740c80-b1a4-11ea-8499-1e2752199100.gif)

It happens in all the directions, not just _left-to-right_.

However, seems to be removing only the belts:
![tunnel-error-2](https://user-images.githubusercontent.com/67031075/85082342-03d17400-b1a5-11ea-87cc-48f453c95a24.gif)

Magos:
I made a fix for the cases displayed here which still leaves some cases possible but involves a minimum of extra performance spend. Not sure if tracing the full path of the belt to check if removal should happen is worth it.

tobspr:
Moved to trello https://trello.com/c/FOD0v6n6/198-improve-smart-tunnel-placement-by-only-removing-a-whole-row-of-belts

PR #238, opened 2020-06-20
Prevents some smart tunnel wrong belt removals (Issue #230).

Magos:
This is a simple fix to prevent the most common instances of smart tunnel placement removing belts it shouldn't. Belt removal is only performed if the final belt points in the same direction as the tunnel exit.
It does not pathfind to make sure the removed belt actually ends at the exit, so wrong removals can still be reproduced but should happen less. 
HJY78GAdbc

### DIFF
--- a/src/js/game/systems/underground_belt.js
+++ b/src/js/game/systems/underground_belt.js
@@ -104,9 +104,13 @@ export class UndergroundBeltSystem extends GameSystemWithFilter {
                 return;
             }
 
-            // Remove any belts between entrance and exit which have the same direction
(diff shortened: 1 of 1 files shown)


tobspr:
Good idea - shouldn't we make a proper solution instead though? What if we limit it so it *only* replaces belts if *all* belts face the same direction as the exit?

Since there is only one use case for the replacement i'd say - and that is replacing an already existing belt part.

tobspr:
Obsolete as 1.1.17 will have this fixed

Issue #240, opened 2020-06-20
Add an option to disable warnings


jimmyaffatigato:
Later in the game, situations frequently arise where it is necessary to cut or delete many buildings at once. In these cases, the game **always** asks for confirmation that requires using the mouse to click OK. It is an annoying speedbump at that point in the game. I noticed that the keybind to delete selected buildings is called "Confirm Mass Delete," but it doesn't dismiss this warning.

**Suggestions:**
- This warning prompt could accept [Enter] as OK and [Esc] as Cancel.
- Holding shift or ctrl when cutting or deleting could skip the warning prompt.
- Add an option under Settings to remove these prompts entirely.
- Rename keybind "Confirm Mass Delete" to "Remove" to be consistent with the hint that appears onscreen when multiple buildings are selected.

tobspr:
This should be a very easy first issue.

Just change "ok:good" to "ok:good:enter" and "cancel:bad" to "cancel:bad:escape" or so

tobspr:
Fixed in 1.1.17

PR #245, opened 2020-06-21
Add option to disable warnings

hexagonhexagon:
Resolve issue #240 by adding a setting to disable the warning dialogs, making the dialogs interactable with keyboard, and changing some localizations slightly.


tobspr:
Nice, thanks!

PR #246, opened 2020-06-22
Updates tier 2 underground belt tutorial icon

jimmyshadow1:
Makes the tutorial icon for the tier 2 underground belt more descriptive by showing off the "mix-and-match" that the description introduces. 

Icon by brimu#5183.


tobspr:
While I appreciate the contribution, I think the mixing and matching can be considered some advanced tactic which doesn't have to be teached.

I would like the tutorials to be simple and on point - this one is pretty advanced

tobspr:
Actually - This seems a good idea after having a look again! :P

Tuxhedoh:
> 
> 
> While I appreciate the contribution, I think the mixing and matching can be considered some advanced tactic which doesn't have to be teached.
> 
> I would like the tutorials to be simple and on point - this one is pretty advanced



> 
> 
> Actually - This seems a good idea after having a look again! :P

Glad you accepted this. I was having a significantly hard time understanding how to layer the belts... until I saw the image.

PR #247, opened 2020-06-22
Change auto save interval to 10 min

Ps7cho:
Due to late game save times taking 10+ seconds taking up 1/6 of in game time. 
When auto save while Quitting is already implemented,  a one minute interval seems excessive.


tobspr:
I will add a setting with the next updates, for the beginning 10 minutes is way too high I feel so we need a setting

PR #249, opened 2020-06-22
I want a swapper building.

isaisstillalive:
Because it is complicated to build with a width of 2 squares.
swapper


hexagonhexagon:
See #26, no chance of this happening anytime soon.

tobspr:
Yup, its way too specialized

isaisstillalive:
OK. Thank you.

PR #252, opened 2020-06-22
Update base-ru.yaml

AngensNofori:
I don't think that Russian " тыс." and others are good. Most of the players do recognize k M B and T as a suffix for large numbers so be it. Otherwise, there was a bug. See in the discord channel for shaking.mp4 attached file.


tobspr:
Manually merged it, thanks!

PR #259, opened 2020-06-22
Polish the Polish translation™ 10

mini-bomba:
So i took a break (6 days to be exact) from translating this game...
In this time there were 2 PRs about changing the Polish translation, so this is why this PR is PtP 10, not 8.

Anyways, in this PR I:
* Added some missing translation that previous translators did not include or were just added (colourblind mode included)
* Made some minor changes to translations made by those new translators. (**_Polishing the Polish translation_** yet again)

This should be ready to merge. And yes, I translated the colourblind mode before the sync commit was pushed.


tobspr:
Nice, thank you! :D

PR #261, opened 2020-06-22
Keep Belt planner mode while dragging if release the shift key

isaisstillalive:
Belt planner is a great feature, but holding down the shift key is very tiring.

My suggestion is that once you enter Belt planner mode, it will continue to work even if you release the shift key.


tobspr:
I feel like this behaviour might be too hard to explain. I'll close this for now until it gets re-requested multiple times

PR #282, opened 2020-06-23
Add 1-to-2 compact balancers.

hexagonhexagon:
As it says in the title. Feel free to improve the art, just making this available if it interests you.
splitter-compact-inverse-split


tobspr:
Looks good but I was actually planning to add that for end game content (So you first unlock the compact merger, then the compact splitter). Still, this is a nice base!

Sandwichs-del:
Actually, I'd prefer to see 1-to-2 compact overflow buildings. 
Balancers are poor choice when supplying rows of buildings.

Vaais:
I'm not a heavy coder so can't add to this, except that I can't count the amount of times I thought "this module that I'm trying to design would be so much better if I had a compact splitter instead of a balancer". Even with a different 1x2 size (1 input anywhere, and 1 output on a short side and 1 on the long side) would open new design possibilities.

Still, (and maybe for that exact reason) I'd implore you to make this item as late game as possible. I love how the game keeps giving you new items to make existing designs better (like a second tunnel, double painter...). It's really one of the most rewarding things about it. A 1x1 splitter, though simple, would be the biggest gamechanger ever, and thus it should be one of the ultimate rewards ;)

tobspr:
Merged it : )

PR #283, opened 2020-06-23
Allow Q-ing on a ore-/shape-patch

Gerdon262:
This simple edit allows for the miner silhouette to show up when pressing q (pipette) on an ore-/shape-patch


src/js/game/hud/parts/building_placer_logic.js
```diff
@@ -244,10 +245,15 @@ export class HUDBuildingPlacerLogic extends BaseHUDPart {
 
         const worldPos = this.root.camera.screenToWorld(mousePosition);
         const tile = worldPos.toTileSpace();
-        const contents = this.root.map.getTileContent(tile);
 
+        const contents = this.root.map.getTileContent(tile);
         if (!contents) {
-            this.currentMetaBuilding.set(null);
+            const tileBelow = this.root.map.getLowerLayerContentXY(tile.x, tile.y);
+            if (tileBelow) {
```
tobspr, line 252: I think it should select the chained extractor here if available, otherwise fine :) (`this.currentVariant` needs to be set)

Gerdon262: This will force people to use the chain extractor when available. I think most users will use the chain extractor after it's available. I haven't reset the variant because of that reason. This way it won't force you to the chain extractor even after you've chosen to use the normal extractor.

tobspr: Why would you use the normal extractor? it has no advantages over the chained one

tobspr:
Will do minor adjustments but looks fine! Thank you!

PR #288, opened 2020-06-24
Fix Building Placer Sound Bug

TristonStuart:
Fixes the sound distortion when using the building placer caused by many buildings trying to play their sounds at once.
This will only play 1 building sound.
Includes checks for if the building was actually placed, so if no buildings were placed then no sound is played and can handle the first buildings not being placed but if at least one is placed it will work.

### DIFF
--- a/src/js/game/hud/parts/building_placer_logic.js
+++ b/src/js/game/hud/parts/building_placer_logic.js
@@ -366,13 +366,16 @@ export class HUDBuildingPlacerLogic extends BaseHUDPart {
     /**
      * Tries to place the current building at the given tile
      * @param {Vector} tile
-     * @param {Object} options
+     * @param {boolean} suppressSound
      */
-    tryPlaceCurrentBuildingAt(tile, options = {}) {
+    tryPlaceCurrentBuildingAt(tile, suppressSound = false) {
         if (this.root.camera.zoomLevel < globalConfig.mapChunkOverviewMinZoom) {
             // Dont allow placing in overview mode
             return;
         }
+
+        // Set sound to true if no options.suppressSound or set to opposite of options.suppressSound.
+        const sound = !suppressSound;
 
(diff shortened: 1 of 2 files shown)


tobspr:
I would change the entire logic and move the playing of sound outside of the game logic into the building_placer_logic.

That should be much cleaner

TristonStuart:
> I would change the entire logic and move the playing of sound outside of the game logic into the building_placer_logic.
> 
> That should be much cleaner

Ya sounds better, I'll add it tomorrow. No idea why the check if failing because of FFmpeg, doesn't make any sense.

TristonStuart:
Ok, I added your suggestion and got the code to a place I like. If you have any more ideas for improvements just let me know.

src/js/game/hud/parts/building_placer_logic.js
```diff
@@ -366,13 +366,17 @@ export class HUDBuildingPlacerLogic extends BaseHUDPart {
...
      */
-    tryPlaceCurrentBuildingAt(tile) {
+    tryPlaceCurrentBuildingAt(tile, suppressSound = false) {
         if (this.root.camera.zoomLevel < globalConfig.mapChunkOverviewMinZoom) {
             // Dont allow placing in overview mode
             return;
         }
 
+        // Set sound to the opposite of suppressSound.
+        const sound = !suppressSound;
```
tobspr, line 378: Why not just add a playSound = true parameter?

TristonStuart:
Requested changes resolved with https://github.com/tobspr/shapez.io/pull/288/commits/02e3448ec653863fc16b4e06f928826e242984b0

tobspr:
I'm sorry but I wanted to do some refactorings anyways, which made this obsolete: https://github.com/tobspr/shapez.io/commit/66eac9346080cc14c6d1c047889726ffeb0324c1

PR #297, opened 2020-06-25
Fix blueprint bugs, take 2: electric boogaloo

hexagonhexagon:
Apparently the first time I tried this, it didn't cover all the necessary cases, so I have taken a different approach and decided to make it so that `updateSurroundingBeltPlacement()` is only called when belts are manually placed rather than when any entity is placed, and if they are placed by blueprint then only the belt cache is updated for that area.

I have also fixed another bug that was described in the discord where the belt blueprint preview didn't match the thing you were copying.

test


hexagonhexagon:
This has been updated to work with the new belt changes.

kubiix:
Please can you check if this case is solved too ?

obrazek

obrazek

obrazek

hexagonhexagon:
Looks like this fix solves that case.

test

kubiix:
Found one more to test

obrazek

tobspr:
Cool! I'll try to get to it shortly, but notice that it might be broken with the wires changes :D

hexagonhexagon:
T\_T

tobspr:
@hexagonhexagon this should be much easier now since the building code is now stored. Should I just close it though?

tobspr:
Outdated, also I won't fix this since it causes other issues (Even if the belt look exactly like the blueprint after placed, this removes the ability to update existing belts)

PR #301, opened 2020-06-25
Fix duplicate waypoints with the same shape key displaying wrong

### LINKED ISSUE #294: Duplicate markers not showing correctly on list view
If I create two markers with the same graphic markup, the graphics show on the map correctly. On the right side of the map, it lists both markers, but one has no graphic.

image

When I click the marker, it takes me to the correct location:

image

Steam and website version both show this issue.

hexagonhexagon:
Fix #294. The reason this happened is because if you call appendElement twice with the same canvas, it will simply move the existing canvas to the new position rather than produce a copy of the canvas. The way to resolve this is to duplicate the canvas before attaching it to the waypoint display.

image


tobspr:
Ah nice, ty!

Issue #284, opened 2020-06-23
Feature request; mirrored cutter


TheTrueSCU:
At one of the levels, you get the counter-clockwise rotate variant.  Could a mirrored cutter variant be added as a level perk?  Some layouts would be more pleasing and space efficient if you could place cutters \\/ instead of just // because a balancer can feed into both cutters directly.

TheTrueSCU:
Can the stacker and the color mixer also have a mirrored version?
It appears the painter and advanced balancers are the only buildings to have mirrors available.

tobspr:
https://trello.com/c/QlYfbg9c/270-allow-mirroring-blueprints-all-buildings

Issue #257, opened 2020-06-22
Feature request: Mirror Blueprints


Jyntax:
It would be nice to have the option to mirror your Blueprints. To quickly build a huge Setup that's maybe fed Objects from the left and to mirror it so that you have a second Setup next to it that can be fed from the right - to double the output quickly without editing the Setup manually.

Best regards,
Jyntax

maweki:
#174

Sandwichs-del:
PR was closed before gerw showed he implemented all missing buildings :(

tobspr:
https://trello.com/c/QlYfbg9c/270-allow-mirroring-blueprints-all-buildings

Issue #174, opened 2020-06-16
Add mirroring/flip to Blueprints.


Jonis326:
It would be a nice feature for blueprinting. Sometimes rotating isn't enough.

maweki:
> It would be a nice feature for blueprinting. Sometimes rotating isn't enough.

That would only work if all components (already) had a mirrored version. The output of the cutter need to be mirrored, as does the orientation of the painter.

lipki:
So, it would be nice to have all the structures in a mirror version.
Which we could choose like the rotation, by pressing [M]

![all_orient](https://user-images.githubusercontent.com/401965/85136771-69bf0980-b240-11ea-9704-8d0601eed86c.png)
(I like to make pictures.)

tobspr:
Appreciate the effort, but I don't think this will happen soon - Closing it thus.

PR #302, opened 2020-06-25
Mirror buildings and blueprints

gerw:
This patch allows to mirror buildings and blueprints, see #174, #257, #284.

The only problem is that it breaks the saves. The `StaticMapEntityComponent` has a new variable in its `schema` and I do not know how to incorporate this into the savegame version logic.

### DIFF
--- a/src/js/core/vector.js
+++ b/src/js/core/vector.js
@@ -520,6 +520,49 @@ export class Vector {
     }
 
     /**
+     * Mirrors this vector
+     * @param {boolean} mirror
+     * @returns {Vector} new vector
+     */
+    mirror(mirror = true) {
+        if (mirror) {
+                return new Vector(-this.x, this.y);
+        }
+        else {
+                return new Vector(this.x, this.y);
+        }
+    }
+
+    /**
+     * Helper method to mirror a direction
+     * @param {enumDirection} direction
+     * @param {boolean} mirror
+     * @returns {enumDirection}
+     */
+    static mirrorDirection(direction, mirror = true) {
+        if (mirror) {
+            switch (direction) {
+                case enumDirection.left: return enumDirection.right;
+                case enumDirection.right: return enumDirection.left;
+            }
+        }
+        return direction;
+    }
+
+    /**
+     * Helper method to mirror an angle
+     * @param {number} angle
+     * @param {boolean} mirror
+     * @returns {number}
+     */
+    static mirrorAngle(angle, mirror = true) {
+        if (mirror && (angle == 90 || angle == 270) ) {
+            return 360 - angle;
+        }
+        return angle;
+    }
+
+    /**
      * Helper method to rotate a direction
      * @param {enumDirection} direction
      * @param {number} angle
--- a/src/js/game/buildings/belt_base.js
+++ b/src/js/game/buildings/belt_base.js
@@ -141,7 +141,7 @@ export class MetaBeltBaseBuilding extends MetaBuilding {
      * @param {string} variant
      * @return {{ rotation: number, rotationVariant: number }}
      */
-    computeOptimalDirectionAndRotationVariantAtTile(root, tile, rotation, variant) {
+    computeOptimalDirectionAndRotationVariantAtTile(root, tile, rotation, mirrored, variant) {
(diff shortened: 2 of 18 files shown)


tobspr:
What about buildings which don't have a mirrored variant?

gerw:
The PR does not care about mirrored variants. It just mirrors the functionality and the sprite of (almost) all buildings. An exception is e.g. the rotater which does not get mirrored.

tobspr:
Uh, that feels kinda hacky. Just mentioning in case you want this PR to get merged actually.

gerw:
Yes, it would be nice if the PR would get merged.

But why do you feel that this is hacky? It is a quite natural process to mirror buildings. You can already rotate buildings and mirroring is quite similar to rotating. Actually, the mirroring is implemented very similar to the rotation of buildings (in the local-to-world and world-to-local transformations).

gerw:
(Just to be sure: It only mirrors the location of the input/output behaviour, but not the behaviour of the item processor. That is, a mirrored factory just produces the vary same shape (and not the mirrored shape). Again, this is the same for rotations: a rotated factory does not produce rotated shapes.)

weasellin:
Hi @gerw 

I agree mirroring blueprints would be cool and convinient.
But like @tobspr said, some factories, ex. cutter, don't have a mirrored variant.
And maby only mirrors the location might yield some unexpected results.

Screen Shot 2020-07-01 at 9 30 56
would become something like
Screen Shot 2020-07-01 at 9 31 03
for example.

tobspr:
Closing since I don't plan to merge it this way, if then I would create a mirrored variant of all buildings

gerw:
@weasellin: No, the building gets mirrored:
image

Here is a mirror of a larger factory:
image



@tobspr: Why is the rotation of buildings 'good', but mirroring is 'bad'?

I don't understand why you would add mirrored variants of all the buildings. This would result in much more lines of code, making it more error-prone. (In fact, you also do not have rotated variants of all the buildings for the same reason) In my opinion, this is much more hacky than my approach (but of course, this is only my opinion). Currently, the compact merger has a horizontal mirror image, while the painter is vertically mirrored, i.e., you need a case distinction if you want to implement mirroring in this way. In my (again) opinion, it would be much cleaner, if there are no mirrored variants at all, but if all buildings can be mirrored. (This also results in a smaller list of variants for each building, making it easier to find the desired building).

Moreover, it should be possible that one does not mirror the sprite of the building, but to use a mirrored variant of the sprite (if it is available).

weasellin:
@gerw oh, I understand your idea now.
Sorry, I haven't look in to the implementation.
However, "Mirroring" as a function, just like "Rotation", does sound simpler and make sense,
at lease for me as a player.

Sandwichs-del:
This is so cool. Can you ping tobspr that you actually implemented missing buildings?

roipoussiere:
@tobspr any chance to get this merged? :slightly_smiling_face:

As a player this feature is very interesting to build compact but complex factories.

heav-4:
i agree. having to manually build mirrored variants of chiral things in my factory is fairly painful and time consuming

schaetzc:
I'd be very happy to see mirrored buildings and hotkeys for mirroring too.

 Mirroring the building horizontally/vertically using dedicated hotkeys (for instance <kbd>H</kbd>/<kbd>V</kbd>) would be so much more comfortable than having to switch through all variants to get a mirrored copy (and then having to rotate that one if you are unlucky). Here is an extreme example: If you unlocked all 3 variants of the painter and would like to switch to a mirrored variant (or back) you may have to press T five times.

flip

As a player, I wouldn't mind if the mirrored buildings were *implemented* as variants. However, I wouldn't want to see the mirrored versions in user interface's variant list. Instead, I would use hotkeys to mirror buildings. The mirroring hotkeys should work for all buildings, even for rotaters.

asdkant:
@tobspr you don't plan on just merging this particular pull request, or is mirroring (I mean from user perspective regardless of underlying implementation) something specifically outside your game design?

As a player, I'd love to have mirror versions of all the buldings that follow the pattern of color mixer, stacker, etc.

PR #304, opened 2020-06-26
Fine tuning for Trad. Chinese translation.

weasellin:
There are two types of tunings:

-  Language fixing, some text are not in Traditional Chinese. So, add the translation.
    - Some text from English to Traditional Chinese.
    - Some text from Simplified Chinese to Traditional Chinese.
- Wording and sentences tunings.
    - These part of tuning might be arguable, because as @YellowSugarHK mensioned the major users for Traditional Chinese are hongkongers and taiwanese. And I might only be able to reconize the common usage in Taiwan (and maybe some of my personal subjective opinions ;) ). So, please help me identify any non-sense changes from the hongkongers' point of view.

### DIFF
--- a/translations/base-zh-TW.yaml
+++ b/translations/base-zh-TW.yaml
@@ -26,7 +26,7 @@
 # Shape：圖形
 # tile：格子
 # Keybind：按鍵設置
-# Menu：主界面
+# Menu：主選單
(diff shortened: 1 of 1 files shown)


tobspr:
Ty!

PR #310, opened 2020-06-26
fix crash due to missing import of STOP_PROPAGATION

jakobhellermann:
can be triggered by pressing the hotkey for a not yet unlocked toolbar item on beta.shapez.io.

crash message


tobspr:
TY!

PR #316, opened 2020-06-27
Fix tunnel incorrect delete bug

mordof:
When trying to place tunnels vertically underneath horizontally run tunnel exits, previously placed tunnel connectors would get removed.

Placed exit at 1# Both exit, and entrance (which were placed previously) at 2# got removed.
image

There was no direction check on the code to identify the connecting tunnel piece, so it was picking up the exit just below the 2# area as it's farthest connector - and then deciding that since that was a viable connector, it didn't need the two tunnel pieces in between.

This adds the direction check in so it doesn't pick up the wrong tunnel connector.


tobspr:
Nice, ty!

PR #325, opened 2020-06-27
press alt key in belt planner to reverse direction

jakobhellermann:
this is done by swapping the start and end tile when computing the preview path.
I renamed the `mouseTile` variable to `endTile` since the the path may start with the mouse now.


tobspr:
Oh cool!

dengr1065:
hah i was waiting for someone to do this

PR #326, opened 2020-06-28
Add "Delivered to storage" to Statistics

isaisstillalive:
Even with the same shape, it is recorded for each storage.
PR_delivered_to_storage

### DIFF
--- a/src/css/ingame_hud/statistics.scss
+++ b/src/css/ingame_hud/statistics.scss
@@ -166,5 +166,28 @@
                 }
             }
         }
+
+        .shape {
+            position: relative;
+
+            button.moveButton {
+                @include S(width, 11px);
+                @include S(height, 11px);
+                background: uiResource("icons/waypoint.png") center center / 95% no-repeat;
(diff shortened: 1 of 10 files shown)


tobspr:
Hey! I appreciate this, but wouldn't it make more sense to show this information when hovering it?

isaisstillalive:
Yes, I would like this information to be displayed when hovering.
But I didn't understand the standard hover information UI, so I tried displaying it on the Statistics screen.

tobspr:
Okay, well I would like to avoid a new category, so it'd be nice to have it as tooltip instead ..

isaisstillalive:
@tobspr 
How about switching between quantity and throughput with a click of storage?
storage_display_throughput

tobspr:
Ah! Yeah thats a cool idea. Looks good, will try to merge soon

tobspr:
I will add a belt reader building soon, which will make this obsolete - Otherwise there would be two ways to measure belt speed

PR #327, opened 2020-06-28
Fix issue where pressing B just after loading crashes the game.

hexagonhexagon:
This is something I have encountered while testing that is infuriating personally. This happens because root hasn't fully initialized yet, so trying to resolve `this.root.gameAnalytics.handleLevelCompleted()` in `onGoalCompleted()` results in an "Can't access property x of undefined" error.


tobspr:
Ty :D

PR #328, opened 2020-06-28
Clear cursor when trying to pipette the same building twice.

hexagonhexagon:
Whenever I find myself using the pipette, I inevitably try to press Q to clear my cursor afterward, only to have it do nothing since there is a building under my cursor, and it just pipetted that instead. This makes it so that you can use Q to clear your cursor, but preserves the functionality of being able to pipette a different building without clearing your cursor.


tobspr:
Hm yeah makes sense. Merging but might have to reconsider this if people don't like it

PR #334, opened 2020-06-29
Serializer/Deserializer

isaisstillalive:
Serializer always outputs in the order of input 1 and 2.
Deserializer always outputs 1 and 2 in that order.
This allows different types of items to be transported on one belt.

PR_serializer

Inspired by 
https://discord.com/channels/706121159405731910/709467254605217793/726912169278439506


isaisstillalive:
Maybe it's better to have 4 slots
PR_serializer2

tobspr:
Uhm well .. this seems like a very specialized building. I currently have no plans on merging this

PR #337, opened 2020-06-29
Bleacher

isaisstillalive:
Convert the Shape to colorless.

It's a convenient but not required building so it works slower (than Painter).
If you care about speed, you should build a circuit without this.

PR_bleacher


tobspr:
I don't currently plan on merging this, but I have some ideas for a similar building.

PR #119, opened 2020-06-12
Add inverse cycling for buildings and variants

SStrippel:
This PR adds inverse cycling functionality for both buildings and variants. 

The default modifier key is "SHIFT".

PR #351, opened 2020-06-30
Reverse cycle when pressing shift key

isaisstillalive:


tobspr:
Duplicate of #119 (to which I didn't get to yet)

Issue #157, opened 2020-06-14
Bug: Shapes increasing in size after a stack and cut


neilCrosswaite:
# Bug
I have listed two problems here about shapes changing size

## Stacker
When you stack a shape that includes a small shape it forces the shape to resize back into the bigger version as seen in this example. This impacts being able to finished level 18.

#### Example
The result:

![image](https://user-images.githubusercontent.com/43375279/84604456-13dc1180-ae8e-11ea-8502-94c67515d502.png)

The expected result (with some help from paint):
![image](https://user-images.githubusercontent.com/43375279/84604577-df1c8a00-ae8e-11ea-8213-0a172a423b8a.png)

## Cutter
I don't know if this is related or a new bug but I came across first and thought might have been a feature.

When you cut a small shape so it is on it's own it returns to the original size. 

#### Second Examples
![image](https://user-images.githubusercontent.com/43375279/84605016-e85b2600-ae91-11ea-93f4-f9dc214c617e.png)

![image](https://user-images.githubusercontent.com/43375279/84604828-b2697200-ae90-11ea-9fe0-c9b3a9175ff4.png)

dengr1065:
stacker: counterintuitive, but technically not a bug. **will be changed soon**
cutter: there's an empty layer behind! it will be brought to bottom one.

tobspr:
What @dengr1065 said. Will soon be obsolete

cyantree:
Will the cutter part be obsolete too? If not, I've prepared a possible fix for it. I can also create a PR if wanted.

*shape_definition.js*
```javascript
    cloneFilteredByQuadrants(includeQuadrants) {
        const newLayers = this.internalCloneLayers();
        let lastNonEmptyLayer = -1;
        for (let layerIndex = 0; layerIndex < newLayers.length; ++layerIndex) {
            const quadrants = newLayers[layerIndex];
            let anyContents = false;
            for (let quadrantIndex = 0; quadrantIndex < 4; ++quadrantIndex) {
                if (includeQuadrants.indexOf(quadrantIndex) < 0) {
                    quadrants[quadrantIndex] = null;
                } else if (quadrants[quadrantIndex]) {
                    lastNonEmptyLayer = layerIndex;
                }
            }
        }

        // Remove top most empty layers which aren't needed anymore
        if (lastNonEmptyLayer !== newLayers.length - 1) {
            newLayers.splice(lastNonEmptyLayer + 1);
        }

[...]

PR #352, opened 2020-06-30
Fix layer handling in cutter

cyantree:
This PR fixes cutting multi layered shapes. Previously all empty layers got dropped which would result in a single second layer quad pushed to the first layer. The fix works by removing only the topmost empty layers.
It's the same fix I already posted in https://github.com/tobspr/shapez.io/issues/157#issuecomment-650397342

### DIFF
--- a/src/js/game/shape_definition.js
+++ b/src/js/game/shape_definition.js
@@ -177,6 +177,11 @@ export class ShapeDefinition extends BasicSerializableObject {
      */
     static isValidShortKeyInternal(key) {
         const sourceLayers = key.split(":");
+
+        if (sourceLayers.length === 0 || sourceLayers.length > 4) {
+            return false;
+        }
+
         let layers = [];
         for (let i = 0; i < sourceLayers.length; ++i) {
             const text = sourceLayers[i];
@@ -220,10 +225,6 @@ export class ShapeDefinition extends BasicSerializableObject {
                 return false;
             }
             layers.push(quads);
-        }
-
-        if (layers.length === 0 || layers.length > 4) {
-            return false;
         }
 
         return true;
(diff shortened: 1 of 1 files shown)


tobspr:
@hexagonhexagon could you approve this?

hexagonhexagon:
A little confused here: what do you mean by approve? In any case, I think lines 218-222 need to be removed from shape_definition.js (says that valid short keys can't have entirely empty layers, which this will override with the new cutting logic) before this is good.

dengr1065:
i don't like idea of leaving full-empty layers at all

cyantree:
In my opinion this depends what's the desired and expected behaviour. When I stumbled upon this behaviour while playing I thought it to be a bug. Also there are multiple issues regarding this. So for me it seems that the expected behaviour would be simply cutting the shape and not transforming any layers.
However if the desired behaviour is lifting shapes from empty layers then maybe another description could help avoid confusion.

@dengr1065 What do you dislike about full-empty layers? Do they complicate things or lead to other problems?

cyantree:
> In any case, I think lines 218-222 need to be removed from shape_definition.js (says that valid short keys can't have entirely empty layers, which this will override with the new cutting logic) before this is good.

Thanks for the catch @hexagonhexagon . If you all think this should be merged I can fix this too. :)

tobspr:
I didn't get to have a look yet but I guess it might make sense - no guarantee though

cyantree:
np, take your time. :)
The fix / changed behaviour could be considered a breaking change for saves as existing setups could need adjustments to function properly again.

cyantree:
@hexagonhexagon I incorporated your suggestion by checking whether the empty layer is allowed at the specific location.

dengr1065:
think like you're stacking and cutting paper pieces, if you cut from top to bottom and from left to right, everything will fall (because of gravity)
also, there would be almost no visible difference between `CuCuCu--:------Cu` and `CuCuCu--:--------:------Cu`

tobspr:
Need to test it

tobspr:
I tested it, but i don't like the behaviour. As @dengr1065 said, I don't like the idea of leaving empty layers. With the redo of the stacking algorithm most of the bugs should now be fixed

cyantree:
Ok. Looking forward to the next release. :)

PR #365, opened 2020-07-01
Fixing Mapping error.

rogeriocamorim:
Due to the : in the text, it was causing some error while compiling.


rogeriocamorim:
more files will be fixed

PR #367, opened 2020-07-01
Initial planner direction based on cursor movement

MizardX:
The orientation of the planner is now considered indeterminate initially. As soon as the cursor is moved at least one tile, the direction gets set based on the direction of movement.

This makes the planner more intuitive, and would lead to less need to flip it with <kbd>R</kbd>.

I was not sure where to place the code, so it ended up in `get currentDirectionLockCorner()`. If you have a better suggestion, I will move it.

Does this need a new option in game settings?


tobspr:
Looks good, definitely gonna check this out!

I think no setting is required for now, as soon as people complain I can still add one

tobspr:
Ty!

tobspr:
Tested, this is such a nice QoL improvement!

PR #372, opened 2020-07-02
Fixed an issue that caused an error when pressing the tab key

isaisstillalive:
Fixed an issue that caused an error when pressing the tab key

It was a combination of the following issues.
* Toolbars on hidden layers are also affected by the tab key.
* If the toolbar has only one item registered, the tab key does not work properly.

### DIFF
--- a/src/js/game/hud/parts/base_toolbar.js
+++ b/src/js/game/hud/parts/base_toolbar.js
@@ -103,14 +103,24 @@ export class HUDBaseToolbar extends BaseHUDPart {
      * Cycles through all buildings
      */
     cycleBuildings() {
+        const visible = this.visibilityCondition();
+        if (!visible) {
+            return;
+        }
+
+        let process = false;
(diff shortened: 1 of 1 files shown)


src/js/game/hud/parts/base_toolbar.js
```diff
@@ -103,15 +103,25 @@ export class HUDBaseToolbar extends BaseHUDPart {
      * Cycles through all buildings
      */
     cycleBuildings() {
+        const visible = this.visibilityCondition();
+        if (!visible) {
+            return;
+        }
+
+        let process = false;
```
tobspr, line 111: Could you rename this to `anyBuildingMatched` or something similar?

isaisstillalive: I changed the variable name to `newBuildingFound`.

tobspr:
Ty!

PR #376, opened 2020-07-02
A suggestion to refine the tryPassOverItem method of ItemEjectorSystem.

isaisstillalive:
The idea is to give slot of ItemAcceptorComponent a type of processor.
Migration is required, but I think that layer judgment will be simple.
Ideally you should have an enum rather than a string.


tobspr:
This definitely looks interesting. I will have a closer look

tobspr:
Closing for now, since I have no plans on merging this rn

PR #377, opened 2020-07-02
Made minor adaptations for consistency or grammar

EnderDoom77:
Raise is transitive, use rise when there's no object.
Made capitalisation more consistent
Allow/enable followed by either: gerund or noun. Use infinitive when the subject (who is allowed/able) is explicitly stated. (more natural)
Commas must be used around non-determining relative clauses.
Added hyphens to counter-clockwise (counterclockwise also valid), user-contributed, and auto-orientation (prefix can't be left alone).
Added a few objects (shapes / belts) to make the text more explicit where objects should naturally be stated.

### DIFF
--- a/translations/base-en.yaml
+++ b/translations/base-en.yaml
@@ -481,7 +481,7 @@ buildings:
     cutter:
         default:
             name: &cutter Cutter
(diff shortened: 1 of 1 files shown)


tobspr:
Thank you! I'm not a native speaker so I appreciate any corrections :)

PR #379, opened 2020-07-02
Update base-ro.yaml

RisenAlex:


tobspr:
Thank you!

RisenAlex:
No problem! Could u accept the next one as well pls, I'm half way done with the translations.

PR #384, opened 2020-07-03
Super minor changes

Jaeyoon07031:
Just found small changes i can make, not really big.
Someone was probably using google translate for the new parts and they could be replaced with better words so thats just what i did.
These changes are really small so you can wait for me to make more changes later on and then merge it.


tobspr:
Ty!

PR #387, opened 2020-07-04
Allow TAB to be keybinded

swtw7466:
It seems to be intended, but I think there seems to be no problems.
Personally, I want to swap the keybinds of `Cycle Variants` and `Cycle Buildings`.


tobspr:
Hm yeah I don't remember why I ignored it ... I guess it should be fine - Thank you! :)

PR #389, opened 2020-07-04
Add first version of Croatian translation.

ivanbratovic:
This is still incomplete, I still have some things left to translate. However, should be good enough for a first version.


tobspr:
Alright, thank you! :)

Issue #391, opened 2020-07-04
implement nicer random piece generator


chocoman:
implement a better random shape generator, that generates symmetric shapes and mixes only similar colors.

chocoman:
I finished implementation of the improved random shape generator today. I have it on my localhost, how can I proceed?

dengr1065:
make sure it's on separate branch and pull request

chocoman:
![shapes_generator](https://user-images.githubusercontent.com/17728621/86517282-092af180-be28-11ea-8273-5129df4bb049.png)
this is what it looks like. These are randomly generated shapes by the new generator.

chocoman:
It seems that I have insufficient access rights:
```
$ git push --set-upstream origin symmetric-piece-generator
ERROR: Permission to tobspr/shapez.io.git denied to chocoman.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

Dimava:
You should push to chocoman/shapez.io.git

Dimava:
i.e. to your own fork

chocoman:
thank you, I created the pull request. Here it is:
https://github.com/tobspr/shapez.io/pull/392
pipeline is currently failing with the same error as master.

tobspr:
Not currently planned

PR #392, opened 2020-07-04
Symmetric piece generator

chocoman:


chocoman:
Travis failed, but it is likely not an issue with this branch, because master fails with the same errors.

tobspr:
Uhm .. could you give a bit more context?

chocoman:
this merge request is an implementation of the suggestion that I had on discord yesterday:

> Suggestion: better looking random shapes for levels 19+. Frankly, the generated shapes are quite ugly, compared to the shapes in previous levels. Producing them correctly feels less rewarding than with the better looking shapes in previous levels.
I think that forcing symmetry (radial or along vertical, horizontal or diagonal axis) might help. Also colors could be picked so that they aesthetically fit together. I would be interested in implementing that, I guess it could be very straight-forward. Shall I do that? Or are there other plans for the late-game levels?

pictures of the shapes generated by the new generator are in
https://github.com/tobspr/shapez.io/issues/391

tobspr:
I'll have a look!

BaleineSanguine:
Looks like a good idea, can you give some screenshots to see the difference ?

tobspr:
This looks cool and is on my todo list

chocoman:
These shapes were automatically generated by this code. See that pieces are either symmetric along vertical, horizontal or diagonal axis, or they are rotationally symmetric. Each piece uses only a small set of colors that fit together.
shapes_generator

Dimava:
I think you should switch from random generation to seeded-random generation based on world seed and target level
See map generation for samples

chocoman:
> I think you should switch from random generation to seeded-random generation based on world seed and target level
> See map generation for samples

Ok, I can do that

tobspr:
Merged with https://github.com/tobspr/shapez.io/commit/4f6d9785c16e83816233be24b427df17fc6ca2f6

PR #395, opened 2020-07-05
Change the suffix for large numbers to use SI in Dutch

Astavie:
Changes made:
 - change 'k' to uppercase 'K' ([source](https://taaladvies.net/taal/advies/vraag/1818/))
 - change 'B' to 'G'

This PR has been made because the English system does not work for Dutch. As `mln.`, `mlj.`, and `bln.` are too long for the game, I suggest using the SI suffixes instead of the English ones.


tobspr:
ty!

PR #396, opened 2020-07-05
Update base-nl.yaml

Astavie:
There was an untranslated bit of text left over. I've also renamed `Kleurenblind-modus` to `Kleurenblindmodus` since that's how it's called in-game.


tobspr:
ty!

Issue #374, opened 2020-07-02
Feature Request: Allow markers to be edited


Xiretza:
Currently, an existing marker can only be zoomed to (left click) or deleted (right click). Making a typo while creating a new marker is quite annoying because the old marker can't be edited, and you have to create a new one from scratch.

Replacing the right-click action with a modal that offers to either delete the marker or edit its name would be great - similar to how factorio does it:

![image](https://user-images.githubusercontent.com/3107142/86381091-342d0e00-bc8d-11ea-89c2-ffc42c900ae6.png)

tobspr:
Already in

PR #397, opened 2020-07-05
Changed "Delete marker" to "Show rename (or delete) marker dialog"

isaisstillalive:
PR_rename_marker
PR_rename_marker2
PR_rename_marker3

#374

### DIFF
--- a/src/js/core/modal_dialog_elements.js
+++ b/src/js/core/modal_dialog_elements.js
@@ -376,11 +376,12 @@ export class DialogWithForm extends Dialog {
             html += formElements[i].getHtml();
         }
 
-        const buttons = [];
+        let buttons = null;
         if (extraButton) {
-            buttons.push(extraButton);
+            buttons = [extraButton, "cancel", confirmButton];
+        } else {
+            buttons = ["cancel", confirmButton];
         }
-        buttons.push("cancel:bad", confirmButton);
         super({
             app,
             title: title,
--- a/src/js/core/modal_dialog_elements.js
+++ b/src/js/core/modal_dialog_elements.js
@@ -365,23 +365,18 @@ export class DialogWithForm extends Dialog {
      * @param {Application} param0.app
      * @param {string} param0.title
      * @param {string} param0.desc
-     * @param {string=} param0.confirmButton
+     * @param {array=} param0.buttons
+     * @param {string=} param0.confirmButtonId
      * @param {string=} param0.extraButton
      * @param {Array<FormElement>} param0.formElements
      */
-    constructor({ app, title, desc, formElements, confirmButton = "ok:good", extraButton = null }) {
+    constructor({ app, title, desc, formElements, buttons = ["cancel", "ok:good"], confirmButtonId = "ok" }) {
         let html = "";
         html += desc + "<br>";
         for (let i = 0; i < formElements.length; ++i) {
             html += formElements[i].getHtml();
         }
 
-        let buttons = null;
-        if (extraButton) {
-            buttons = [extraButton, "cancel", confirmButton];
-        } else {
-            buttons = ["cancel", confirmButton];
-        }
         super({
             app,
             title: title,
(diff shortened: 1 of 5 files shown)


src/js/core/modal_dialog_elements.js
```diff
@@ -366,20 +366,27 @@ export class DialogWithForm extends Dialog {
...
         for (let i = 0; i < formElements.length; ++i) {
             html += formElements[i].getHtml();
         }
 
+        let buttons = null;
+        if (extraButton) {
+            buttons = [extraButton, "cancel", confirmButton];
+        } else {
+            buttons = ["cancel", confirmButton];
+        }
```
Xiretza, line 384: `cancel:bad` became just `cancel` here

translations/base-en.yaml
```diff
@@ -266,6 +266,10 @@ dialogs:
         title: New Marker
         desc: Give it a meaningful name, you can also include a <strong>short key</strong> of a shape (Which you can generate <a href="https://viewer.shapez.io" target="_blank">here</a>)
 
+    editMarker:
+        title: Edit Marker
+        desc: Give it a new meaningful name, you can also include a <strong>short key</strong> of a shape (Which you can generate <a href="https://viewer.shapez.io" target="_blank">here</a>)
```
tobspr, line 271: Why duplicate the translation key here? :) You can just use the same one

src/js/core/modal_dialog_elements.js
```diff
@@ -366,20 +366,27 @@ export class DialogWithForm extends Dialog {
      * @param {string} param0.title
      * @param {string} param0.desc
      * @param {string=} param0.confirmButton
+     * @param {string=} param0.extraButton
      * @param {Array<FormElement>} param0.formElements
      */
-    constructor({ app, title, desc, formElements, confirmButton = "ok:good" }) {
+    constructor({ app, title, desc, formElements, confirmButton = "ok:good", extraButton = null }) {
```
tobspr, line 372: I don't like extraButton - Why not just make the buttons itself a parameter? E.g. `buttons = ["cancel", "ok:good"]` that way we could also get rid of `confirmButton`

src/js/game/hud/parts/waypoints.js
```diff
@@ -295,6 +284,59 @@ export class HUDWaypoints extends BaseHUDPart {
         this.rerenderWaypointList();
     }
 
+    /**
+     * Requests to edit a marker.
+     * @param {Waypoint} waypoint
+     */
+    requestEditMarker(waypoint) {
```
tobspr, line 291: Could you change it so that the same method for creating and editing a marker is used?

src/js/game/hud/parts/waypoints.js
```diff
@@ -304,6 +346,23 @@ export class HUDWaypoints extends BaseHUDPart {
         }
     }
 
+    /**
+     * Sort waypoints by name
+     */
+    sort() {
```
tobspr, line 352: I like that refactoring, but could you choose a bit longer name? I.e. `sortWaypoints`. Just trying to avoid builtin method names

src/js/core/modal_dialog_elements.js
isaisstillalive: https://github.com/tobspr/shapez.io/blob/2722366d686fd551a8edceebafc8e78c1e40e976/src/js/core/modal_dialog_elements.js#L373
I agree

translations/base-en.yaml
isaisstillalive: https://github.com/tobspr/shapez.io/blob/2722366d686fd551a8edceebafc8e78c1e40e976/translations/base-en.yaml#L283-L286
Prepared another title using the same translation key

src/js/game/hud/parts/waypoints.js
isaisstillalive: https://github.com/tobspr/shapez.io/blob/2722366d686fd551a8edceebafc8e78c1e40e976/src/js/game/hud/parts/waypoints.js#L336

I agree

src/js/game/hud/parts/waypoints.js
isaisstillalive: https://github.com/tobspr/shapez.io/blob/2722366d686fd551a8edceebafc8e78c1e40e976/src/js/game/hud/parts/waypoints.js#L224-L227
How about this?

tobspr:
Looks good - thank you!

PR #403, opened 2020-07-06
Fix preferred variant not getting saved when clicking on the hud

Danacus:
I noticed that while pressing `T` to cycle variants, the last variant was saved, but when clicking on one of them it would not save.
The problem was that the click event handler in `building_placer.js` was only setting `currentVariant` without updating `preferredVariants`. 

I'm not sure if this fix could potentially cause other issues, because I'm not checking if `metaBuilding` exists.


tobspr:
Thank you!

PR #406, opened 2020-07-06
base-it.yaml

RedboomITA:
I've just added some more translations from the English to the Italian. I wish that u will appreciate.


tobspr:
Thank you!

PR #410, opened 2020-07-07
Update colors.js - Fix black + white = gray and not white

MrYawnie:
Mixing black and white now correctly create gray paint, and not white anymore

image


tobspr:
Alreadx fixed myself :P

PR #416, opened 2020-07-07
Add decimal separator option in translation files

### LINKED ISSUE #398: Option to change the decimal point to a comma in translation
Hi!

I already mentioned this on Discord, but I'll also open an issue about it here.

Not all languages write numbers the same way. You already allowed us to change the thousands separator, but there is one more separator that needs to be made mutable: the decimal point.

In English, the above number (`1.5k`) is spelt correctly. But not so much in Dutch and some other languages. Here, we write it `1,5K`, with a comma. I'd love to open a pull request to edit this in the translation file, but there doesn't seem to be a key to change it.

Astavie:
Fixes #398


Astavie:
Well, I can't seem to make the linter happy, hope that isn't too bad

tobspr:
Thank you! Nice pr

PR #417, opened 2020-07-07
Update base-pl.yaml

W0LFB0MB:
Added a few English to Polish translations that were not yet added.
Also switched 'number' to 'numeral' on line 357 as the word number is incorrect when referring to roman numerals.


tobspr:
Ty!

PR #420, opened 2020-07-07
Update base-nl.yaml

Ryoncai:
Translated new content and minor improvements


Ryoncai:
I reverted the large number suffix because this is much more intuitive in my opinion.

tobspr:
okay :D thank you!

Astavie:
Why change G back to B? This isn't at all intuitive to me

PR #422, opened 2020-07-08
Fixed that the processor output interval was slower than the theoretical value

isaisstillalive:
The remainder of the calculation of the output interval of the processor is added to the next output.
Normally, the next process of the processor should be running faster by that time.
This is equivalent to an extractor referencing in-game time.

PR_processor_speed


dengr1065:
A more up-to-date and complete solution is available on #1364. Not sure if that's going to get accepted either.

PR #431, opened 2020-07-09
Update base-fr.yaml

Zilounay:
Added corrections to french in the first 200 lines only.

The original translation had a few mistakes. I will read on the rest of the file later, but at least the store page is now gramatically correct.

The phrasing sometimes seems a bit odd or unnatural, but I don't wanna interfere too much.


tobspr:
Ty!

PR #423, opened 2020-07-08
Update base-no.yaml

NiceManiac:
Translated the Steam store page completely to Norwegian

PR #435, opened 2020-07-09
Update base-no.yaml

Rotekoppen:
La til en oversettelse for Steampage.longtext og steampage.discordlink

Usikker på linje 40 der jeg fjernet et helt ord

Added a translation for Steampage.longtext and steampage.discordlink

Unsure about line 40 where I removed a whole word


Rotekoppen:
Found this one that does exactly the same:
https://github.com/tobspr/shapez.io/pull/423

PR #437, opened 2020-07-10
Change the processing time of ItemProcessor so that it only affects the interval and not the speed.

isaisstillalive:
If ItemProcessor was faster than Belt, it was moving faster than the belt alone.
I think the speed of ItemProcessor is waiting time, not moving speed.

### ItemProcessor > Belt
Current
PR_beltspeed1
New
PR_beltspeed2

### ItemProcessor < Belt
Current
PR_beltspeed3
New
PR_beltspeed4


tobspr:
uh, that looks weird. need to take a closer look

PR #449, opened 2020-07-15
Fixed Floating point error in processor output interval

isaisstillalive:
Just adding quantizeFloat had the same effect as #422

### DIFF
--- a/src/js/game/systems/item_ejector.js
+++ b/src/js/game/systems/item_ejector.js
@@ -8,6 +8,7 @@ import { ItemEjectorComponent } from "../components/item_ejector";
 import { Entity } from "../entity";
 import { GameSystemWithFilter } from "../game_system_with_filter";
 import { enumLayer } from "../root";
+import { quantizeFloat } from "../../core/utils";
 
 const logger = createLogger("systems/ejector");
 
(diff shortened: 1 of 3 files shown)


src/js/game/systems/item_processor.js
```diff
@@ -31,7 +32,7 @@ export class ItemProcessorSystem extends GameSystemWithFilter {
 
             // Check if we have any finished items we can eject
             if (
-                processorComp.secondsUntilEject === 0 && // it was processed in time
+                quantizeFloat(processorComp.secondsUntilEject) == 0 && // it was processed in time
```
Moppler, line 35: Why remove the strict check? Under the hood `quantizeFloat` performs a `Math.round()` operation which has a return type of `number`. This matches the data type of `processorComp.secondsUntilEject` so the struck check should still work.

Moppler: Same question for your changes to `underground_belt.js` both changes arr to properties with a `number` data type.

src/js/game/systems/item_ejector.js
```diff
@@ -225,7 +226,7 @@ export class ItemEjectorSystem extends GameSystemWithFilter {
                 );
 
                 // Check if we are still in the process of ejecting, can't proceed then
-                if (sourceSlot.progress < 1.0) {
+                if (quantizeFloat(sourceSlot.progress) < 1.0) {
```
tobspr, line 229: Just add an epsilon here, not?

PR #451, opened 2020-07-15
Feature: Counter Building

### LINKED ISSUE #448: Feature Request: Counter
A component which i assume would be relatively simple but very neat for the number freaks like me:
A small 1x1 building that you can put over a conveyor belt which displays the amount of items of a specific type which pass through this road each minute. Could be detailed showing all item types, or only one item type at once... but i would love it :)

Moppler:
## User Story

As a player in the later stages of the current story mode, I'm having trouble optimising my factories. As the belt speeds get to 20 (items per second) and above, It is difficult to visually judge output at different points of my shape production. I would like a game object that allows me to track the performance of my factories. It should be simple to use and provide me with a live readout of the items per second passing through my belt.

## Screenshot
Screenshot 2020-07-15 at 16 40 17

## Acceptance Criteria

- Fits in with the existing aesthetic of the game
- Can be used to replace a belt
- Displays a reading that is as accurate as possible
- Works with shapes and colours

## Limitations

- Does not work with corners, all items must pass from bottom to top of the building

Fixes #448


src/js/game/buildings/counter.js
```diff
@@ -0,0 +1,68 @@
...
+        const beltSpeed = root.hubGoals.getBeltBaseSpeed("regular");
+        return beltSpeed >= 20;
+    }
+
+    /**
+     * Creates the entity at the given location
+     * @param {Entity} entity
+     */
+    setupEntityComponents(entity) {
+        entity.addComponent(new ItemCounterComponent());
```
Moppler, line 50: Theoretically, this component could be added to existing buildings too, providing them with the same counting functionality. Could be a good expansion to the feature down the line.

Arescet:
Would love this in the game, even if it's buried under some debug or accessibility setting!

Moppler:
Travis build is failing on master branch with the same reason.

tobspr:
Well, the idea is good, but I don't want to add more buildings in the toolbar. I first need to figure out a good way to add more buildings without polluting the toolbar

MrYawnie:
Add a single "utilities" entry into toolbar. 

> Well, the idea is good, but I don't want to add more buildings in the toolbar. I first need to figure out a good way to add more buildings without polluting the toolbar

Add a single "utilities" item in the toolbar. Then switching between those utility buildings with `T`, just like switching between building variations.

tobspr:
I think instead the toolbar should be configurable, or maybe with a misc tab. I'll have to figure sth out first

MrYawnie:
> I think instead the toolbar should be configurable, or maybe with a misc tab. I'll have to figure sth out first

Configurable toolbar would work a well. And yeah, when I said "utilities", I meant "misc" as well :) Similar idea.

Edited my last message to mention that the color inverter could be added as a variation of color mixer, as both of them deal with colors. That would clean up the toolbar a bit as well. Not sure if you saw the edit.

tobspr:
I think none of the buildings until level 18 (I.e. the current buildings) should get merged since they all are frequently used. I think instead the advanced buildings should be in a different tab

Dimava:
I propose a change in avegage calculation method
https://github.com/Dimava/shapez.io/commit/408212a6de568f39b9c5d38c631f637e10e189af#diff-3ecca86b3e6f6c1866d9b810daa7fba8
(sorry, can't properly test it now)

tobspr:
I'll add a counter building soon, so I'll close this, I'll definitely take inspiration by the artwork though!

PR #458, opened 2020-07-17
Optimized some UI layers for fewer paints on interaction

Djspaceg:
I noticed while looking through the code there were a few places where optimizations weren't happening the way they were intended. This leads to unnecessary paints and slowdowns when the cursor passed over elements.

Changes:
* Several places have had will-change added to support GPU acceleration/offloading
* `transform-origin`, with updated values to match visually, was used instead of an additional transform of the pinned shape for a smoother transition
* Refactored the hover background color change for the toolbar to be more performant
* Converted several awkward time units from seconds to milliseconds for better readability. Ex: `0.05s` -> `50ms`
* **Optional:** A "frosted glass" effect was applied to the key-guide layer. This can be taken out easily if it's not wanted. It's just an idea.

### DIFF
--- a/src/css/ingame_hud/buildings_toolbar.scss
+++ b/src/css/ingame_hud/buildings_toolbar.scss
@@ -4,17 +4,18 @@
     left: 50%;
     transform: translateX(-50%);
 
+    // NOTE: This flex rule may not be necessary. Need to find out intent.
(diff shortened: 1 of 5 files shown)


tobspr:
Thanks, highly appreciated!

PR #460, opened 2020-07-18
Update base-kor.yaml

xbfld:
Translate updated stuff, correct misspells and errors, but left controversial words behind (maybe issuing later)


tobspr:
Ty!

PR #468, opened 2020-07-19
Validate translations

Moppler:
As there are a lot of long translation strings, and much of the translation work is done by non-technical contributors, I have set the line-length rule to warn only with a much higher maximum (default 80).

I also cleaned up the trailing whitespace. This was the minimum required to ensure that the validator passes linting checks.


tobspr:
Awesome, thank you! That really helps

PR #470, opened 2020-07-19
Extend the pipette hack to work with belts and wires.

Moppler:
Extended the pipette building check hack to understand the difference between belts and wires. Ideally, we would replace the hack with a proper implementation but that's going to require a re-work of some core code. Outside the scope of a bug fix.

Fixes Bug: https://trello.com/c/oN4qFsiK/263-q-on-belts-doesnt-work-anymore


tobspr:
Ty!

PR #478, opened 2020-07-21
Update browser-sync from v2.24.6 to v2.26.10

bjornstar:
Also it looks like `browser-sync` is only used by the `builder` so I removed it as a dependency of `shapez.io`


tobspr:
ty!

PR #481, opened 2020-07-22
Add TSLint to github actions, fix existing type errors

bjornstar:
It looks like there is a `tslint` job which can be run by CI. Adding it to the Github CI Workflow was easy, fixing the existing type errors was fairly straightforward as well.

## Summary of type fixes
- Add the `DOM` and `ES2018` libs to `tsconfig.json`
- Remove references to the contest which is no longer implemented.
- `global.d.ts` had definitions for Element which conflicted with the typescript definition
- Some instances of querySelector needing to be cast are suboptimal, maybe there's a better approach here

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,4 +47,7 @@ jobs:
 
             - name: TSLint
               run: |
+                  cd gulp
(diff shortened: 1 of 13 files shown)


tobspr:
Awesome! Really nice contribution

PR #486, opened 2020-07-23
[Perf] Avoid iterating over keys to generate assert message

bjornstar:
I did a quick perf check and this line stuck out. Even though the assert is not triggering, it is iterating over all the keys to build the string.

Before:
shapez-perf-before

After:
shapez-perf-after

Profile duration was approximately 15 seconds realtime.

### DIFF
--- a/src/js/core/sprites.js
+++ b/src/js/core/sprites.js
@@ -159,7 +159,7 @@ export class AtlasSprite extends BaseSprite {
         const link = this.linksByResolution[scale];
 
         if (!link) {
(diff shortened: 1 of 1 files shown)


tobspr:
Well its mostly a dev thing -> It is definitely useful to see which links haven't been loaded. (Asserts are compiled out in prod builds)

What I'd suggest instead:

```js
if (!link) {
  assert(false, "Link not known" + ... + ...)
}
```

tobspr:
ty!

PR #487, opened 2020-07-23
Explicitly list yarn as a devDependency

bjornstar:
This way, if people aren't using `yarn`, most commands will still work.


tobspr:
Ty!

PR #493, opened 2020-07-25
Update simplified Chinese translation 

chaow108:
Translated some missed texts on the levels that I already played.


tobspr:
Ty!

PR #502, opened 2020-07-27
 Improve types for web workers

bjornstar:
We can leverage the `WebWorker` types for the `webworkers` directory.

I couldn't help touching the `animation_frame` to remove a branch that would only occur on the first frame. If you would prefer to leave it as is, I'd be happy to back that out.

### DIFF
--- a/src/js/core/animation_frame.js
+++ b/src/js/core/animation_frame.js
@@ -1,6 +1,6 @@
 import { Signal } from "./signal";
 
-import BackgroundAnimationFrameEmitterWorker from "worker-loader!../webworkers/background_animation_frame_emittter.worker";
+import BackgroundAnimationFrameEmitterWorker from "worker-loader?inline=true&fallback=false!../webworkers/background_animation_frame_emittter.worker";
 
 import { createLogger } from "./logging";
 const logger = createLogger("animation_frame");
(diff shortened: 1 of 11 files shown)


src/js/core/animation_frame.js
```diff
@@ -1,7 +1,6 @@
 import { Signal } from "./signal";
 
-// @ts-ignore
-import BackgroundAnimationFrameEmitterWorker from "../webworkers/background_animation_frame_emittter.worker";
+import BackgroundAnimationFrameEmitterWorker from "worker-loader!../webworkers/background_animation_frame_emittter.worker";
```
tobspr, line 3: I think you need to pass the `fallback` and `inline` options here then, not?

src/js/core/async_compression.js
```diff
@@ -1,7 +1,6 @@
-// @ts-ignore
-import CompressionWorker from "../webworkers/compression.worker";
+import CompressionWorker from "worker-loader!../webworkers/compression.worker";
```
tobspr, line 1: Same here

src/js/core/animation_frame.js
bjornstar: Sorry about that, for some reason I read their docs as inline being the default.

bjornstar:
I cherry-picked my fix for the trailing spaces from #504

tobspr:
ty!

PR #517, opened 2020-07-29
Add support for a 75hz simulation target

moonheart08:
Entirely untested, I just wanted to avoid having it sit as an issue forever and as such just quickly added it with the GitHub editor. Someone actually check it please. :sweat_smile:


tobspr:
Ty!

PR #526, opened 2020-08-02
Remove unused variables

bjornstar:
I noticed a bunch of unused variables and unused exports especially in `core/utils`

I also tried to use more typedef imports instead of real imports to avoid having unused variables as well.

There's lots more to do, but I'd like to get your opinion on this before I do more.

### DIFF
--- a/src/js/core/atlas_definitions.js
+++ b/src/js/core/atlas_definitions.js
@@ -1,20 +1,38 @@
 /**
+ * @typedef {{ w: number, h: number }} Size
+ * @typedef {{ x: number, y: number }} Position
  * @typedef {{
- *   frame: { x: number, y: number, w: number, h: number },
- *   rotated: false,
- *   spriteSourceSize: { x: number, y: number, w: number, h: number },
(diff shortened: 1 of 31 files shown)


src/js/core/atlas_definitions.js
```diff
@@ -1,27 +1,46 @@
...
+ * @typedef {{ x: number, y: number }} Position
  * @typedef {{
- *   frame: { x: number, y: number, w: number, h: number },
- *   rotated: false,
- *   spriteSourceSize: { x: number, y: number, w: number, h: number },
- *   sourceSize: { w: number, h: number},
- *   trimmed: true
+ *   frame: Position & Size,
+ *   rotated: boolean,
+ *   spriteSourceSize: Position & Size,
```
tobspr, line 7: Oh wow, thats handy!

tobspr:
This is definitely helpful! If you want to go ahead, sure - Although for new code I will probably stick to the old style, because then vscode automatically generates the imports for me :)

PR #529, opened 2020-08-03
Replace enumDirection with string literal Direction

bjornstar:
In practice, TypeScript's string enums end up making code more bloated without any of the benefits.

We can replace the string enum type `enumDirection` with a string literal type `Direction` without much hassle.

I also went ahead and created an `Angle` type to enforce consistent usage throughout the codebase. All of the math used for rotating and inverting are implemented using maps.

I've done the colors as well on a separate branch. If you're okay with this approach I'll apply it for the rest of the enums.


bjornstar:

PR #530, opened 2020-08-03
Replace enumColors with string literals

bjornstar:
This was a bit quicker to do than removing enumDirections (see #529)

If you're okay with removing enums, I'll do this for the rest of them.


dengr1065:
Hm, that breaks shapebot, but it's alright. Just need to rewrite whole game in TS, then it will be really ok :slightly_smiling_face:

bjornstar:

PR #531, opened 2020-08-03
Remove enumLayer and enumItemType

bjornstar:
It seems that `Layer` and `ItemType` were linked so I did them both at the same time.


bjornstar:

PR #535, opened 2020-08-04
Add sideways tunnels

Squornshellous-Beta:
I wanted these and was curious to see if they were possible (turns out they are), and several other people have asked for them just in the last few days, so I thought I may as well put in a request.


tobspr:
Cool idea! But currently I have no plans on integrating it :)

PR #541, opened 2020-08-05
Combined enum removal

bjornstar:
I realized that there would be a lot of conflicts if you wanted to go ahead with the enum removal so I combined them all on this branch. This can replace #529, #530, and #531.


tobspr:
I don't like the changes of #529, what's the reason for that? :)

bjornstar:
I'll split these up into separate PRs it will be easier to review that way.

PR #544, opened 2020-08-06
inject KeybindingsState into SettingsState

Dimava:


tobspr:
What does this do? Show keybindings on the settings page?

Dimava:
This seamlessly integrates Keybindings screen into Settings screen
image

Dimava:
You may check keybing settings on https://dimava.github.io/shapez/modZ/ to see how that looks on live.

tobspr:
You are not changing into the "new" settings state though, so this will error, not?

Dimava:
I replace Settings state with combined state (made as KeybindingsState extends SettingsState).
So instead of Setting state there is the Settings (+keybindings) state where both pieces are on at once.
It enables both of their interactivity, and sharing their html space
Html makes one shared element, with one of its children made for inserting other settings and another child for inserting keybindings

Dimava:
The new state is still called "settings" and completely replaces old "settings" state so that does not causes errors
As for "does it work" you may check https://dimava.github.io/shapez/modZ/ where it does work perfectly

tobspr:
Hm, I like the idea, but I think instead I should refactor the settings state since this solution is kinda hacky and I don't like that

tobspr:
Closing since for merging, it would first have to get refactored

PR #549, opened 2020-08-07
Replace enumLayer and enumItemType with string literals

bjornstar:
I'll try this again with smaller batches.

I'd like to find out if you're interested in replacing enums with string literals.

### DIFF
--- a/src/js/game/base_item.js
+++ b/src/js/game/base_item.js
@@ -1,12 +1,8 @@
 import { DrawParameters } from "../core/draw_parameters";
 import { BasicSerializableObject } from "../savegame/serialization";
 
-/** @enum {string} */
-export const enumItemType = {
-    shape: "shape",
-    color: "color",
-    boolean: "boolean",
-};
+/** @type {ItemType[]} **/
+export const itemTypes = ["shape", "color", "boolean"];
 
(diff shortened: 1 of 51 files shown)


tobspr:
Again, what's the advantage here? :)

bjornstar:
There are several advantages:
1) We don't have to import an object from another file to reference. We can use the word `"shape"` or `"color"` which we were going to do anyway.

2) In vscode intellisense will list the acceptable values like so:
shapez-enum-2
instead of looking like this:
shapez-enum-1

3) Enum strings require that both the key and the value be the same for reverse lookup purposes so you're, at a minimum, doubling the size of the string. With just the changes from this branch we lose 10kb from the bundle size.

tobspr:
Okay, so well I think this is a "nice to have" . But I think I broke your entire PR through my changes (Sorry for that!). I think splitting it up into smaller PR's might work better? Especially in the next days a lot of changes are coming in. And sorry for taking so long to respond!

tobspr:
Merged in:
https://github.com/tobspr/shapez.io/commit/6a447e3583a59cc3d1a7108fee053736db2f864d

(I manually rebased it)

PR #550, opened 2020-08-07
Fix types for v1000->v1001 migration

### LINKED ISSUE #514: [CI] TSLint failing
```
##[error]savegame/schemas/1001.js(49,32): error TS2339: Property 'spriteKey' does not exist on type 'StaticMapEntityComponent'.
##[error]savegame/schemas/1001.js(50,32): error TS2551: Property 'blueprintSpriteKey' does not exist on type 'StaticMapEntityComponent'. Did you mean 'getBlueprintSprite'?
##[error]savegame/schemas/1001.js(50,64): error TS2339: Property 'spriteKey' does not exist on type 'StaticMapEntityComponent'.
##[error]savegame/schemas/1001.js(56,36): error TS2551: Property 'blueprintSpriteKey' does not exist on type 'StaticMapEntityComponent'. Did you mean 'getBlueprintSprite'?
##[error]savegame/schemas/1001.js(59,36): error TS2551: Property 'blueprintSpriteKey' does not exist on type 'StaticMapEntityComponent'. Did you mean 'getBlueprintSprite'?
```

https://github.com/tobspr/shapez.io/commit/bf2eee908fedb84dbbabd359a200c446020a340e appears to break the migration from 1000 to 1001.

I don't know enough about the migrations to fix it, but I can raise a PR to at least get CI green again.

asquared31415:
The change to use an ID based system for buildings (bf2eee908fedb84dbbabd359a200c446020a340e) broke the v1000->v1001 migration by making the StaticMapEntityComponent type incompatible with the correspinding type in older saves.  Instead, we can mock the type of the old StaticMapEntityComponent and use it in the migration.  A `@ts-ignore` is required to suppress the warning that the types do not match, but since the code will only be run on old data, the types should always match.

This resolves #514 in a way that keeps type safety.

### DIFF
--- a/src/js/savegame/schemas/1001.js
+++ b/src/js/savegame/schemas/1001.js
@@ -1,6 +1,7 @@
 import { SavegameInterface_V1000 } from "./1000.js";
 import { createLogger } from "../../core/logging.js";
 import { T } from "../../translations.js";
+import { TypeVector, TypeNumber, TypeString, TypeNullable } from "../serialization_data_types.js";
 
 const schema = require("./1001.json");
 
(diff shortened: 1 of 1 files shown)


tobspr:
Oh well, thanks for the effort! A simple `@ts-ignore` would have done it , too :) But a proper solution is always good

PR #552, opened 2020-08-09
Replace enumDirection with string literal Direction

bjornstar:
This also introduces `Angle` which is `0 | 90 | 180 | 270` since everything in the game uses right angles.

I ran into a bug with `inverseAngleMap` where it only works if the vertical directions are not inverted. I'd love some help tracking down how to fix that. I ended up using `clockwiseAngleMap[clockwiseAngleMap[rotation]]` in places that functioned correctly and left the `inverseAngleMap` usage where that behavior is required.

Earlier you said you didn't like this change, but I'd like to hear what you don't like about it.

### DIFF
--- a/src/js/game/blueprint.js
+++ b/src/js/game/blueprint.js
@@ -1,7 +1,6 @@
 import { DrawParameters } from "../core/draw_parameters";
-import { Loader } from "../core/loader";
 import { createLogger } from "../core/logging";
-import { Vector } from "../core/vector";
+import { clockwiseAngleMap, counterClockwiseAngleMap, Vector } from "../core/vector";
 import { Entity } from "./entity";
 import { GameRoot } from "./root";
 import { findNiceIntegerValue } from "../core/utils";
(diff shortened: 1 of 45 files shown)


tobspr:
I'll have a look, it will take some time until I can get to it tho

tobspr:
See my comment on the other PR :) I know it's already just one enum though .. Hm maybe this is not the right time to change it? Since the wires update adds a lot of structural changes

src/js/game/blueprint.js
```diff
@@ -101,42 +100,22 @@ export class Blueprint {
...
     rotateCw() {
-        for (let i = 0; i < this.entities.length; ++i) {
-            const entity = this.entities[i];
-            const staticComp = entity.components.StaticMapEntity;
-
-            staticComp.rotation = (staticComp.rotation + 90) % 360;
-            staticComp.originalRotation = (staticComp.originalRotation + 90) % 360;
-            staticComp.origin = staticComp.origin.rotateFastMultipleOf90(90);
-        }
+        this.entities.forEach(({ components: { StaticMapEntity } }) => {
```
tobspr, line 103: I still don't like forEach :P

PR #556, opened 2020-08-10
Fix sendtoApi uncaught Promise rejection

bjornstar:
Using reject here is uncaught, make this a single promise and resolve/reject accordingly

### DIFF
--- a/src/js/platform/browser/game_analytics.js
+++ b/src/js/platform/browser/game_analytics.js
@@ -85,10 +85,9 @@ export class ShapezGameAnalytics extends GameAnalyticsInterface {
      * @returns {Promise<any>}
      */
     sendToApi(endpoint, data) {
-        return Promise.race([
-            new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


src/js/platform/browser/game_analytics.js
```diff
@@ -103,13 +102,15 @@ export class ShapezGameAnalytics extends GameAnalyticsInterface {
...
-                        throw new Error("Fetch error: Bad status " + res.status);
+                        reject("Fetch error: Bad status " + res.status);
+                    } else {
+                        resolve(res.json());
                     }
-                    return res;
                 })
-                .then(res => res.json()),
-        ]);
+                .catch(reject);
```
tobspr, line 112: If the fetch fails, then the timeout will not be cleared - I'm not sure if thats good practice?

bjornstar: Reject being called a second time doesn't do anything, but I have added a `clearTimeout` and fixed the resolve.

tobspr:
Ty!

PR #558, opened 2020-08-12
Arrange base-de.yaml so it matches base-en.yaml

DerFeldspatz:
I noticed that german and english translation are in a different order so I fixed this. There should be no changes in the existing translations.


anno9474:
For me, it looks clean 👍 
Just compared it with the base-en.yaml :D

DerFeldspatz:
Thanks for double checking :)

tobspr:
Danke!

PR #565, opened 2020-08-13
Replace enumSubShape with SubShape

bjornstar:


tobspr:
See my other comments :)

PR #567, opened 2020-08-14
Use yaml-loader instead of relying on gulp to build the translations

bjornstar:
I also added type definitions for your language files. They don't enforce that your yaml files contain all the fields, but they will help you reference entries that you expect to exist.


tobspr:
First of all, nice pr! I learned a lot :P

I already get autocompletion in vscode though:
image

So it seems to be a big effort maintaining the type definitions for
1. Adding a new language
And
2. Adding new translation keys

What would be the advantage? From what I can see, at least tslint can now check if a referenced translation is supposed to exist

bjornstar:
Currently, in order to run `tslint` on the project we have to execute a gulp task before hand.

This would make it so that we don't have to do that step.

I added the types because we wouldn't be requiring json files anymore and they wouldn't be automatically discovered from the files.

PR #571, opened 2020-08-16
[core/rectangle] Remove unused methods

bjornstar:


tobspr:
Hm, I would prefer if we would keep those - I like that the rectangle class has a lot of utility methods (Just like the vector) and thus already has a method for everything.

The reason they are not used is, is that the code is copied from yorg.io 3. So all those methods actually had a use case. I'd be fine about removing the very special methods like `getMinStartTile`,`extendBySquare` and `getAroundPointsRotated` though

bjornstar:
Okay, I've reduced the changes down to just those three methods.

tobspr:
Thank you!

PR #575, opened 2020-08-18
Reduce boilerplate of BooleanItem 

bjornstar:
I'm sure you had lots of plans for all of this, but I thought I might show you how it could look if you used a type guard.

### DIFF
--- a/src/js/game/items/boolean_item.js
+++ b/src/js/game/items/boolean_item.js
@@ -6,6 +6,12 @@ import { globalConfig } from "../../core/config";
 
 /** @typedef {0 | 1} Bit **/
 
+/** @type {Record<Bit, string>} **/
+const bitSpriteMap = {
+    0: "sprites/wires/boolean_false.png",
+    1: "sprites/wires/boolean_true.png",
+};
+
 export class BooleanItem extends BaseItem {
     static getId() {
         return "boolean_item";
@@ -50,8 +56,7 @@ export class BooleanItem extends BaseItem {
      * @param {DrawParameters} parameters
      */
     drawItemCenteredImpl(x, y, parameters, diameter = globalConfig.defaultItemDiameter) {
-        const value = Boolean(this.value).toString();
-        const sprite = Loader.getSprite(`sprites/wires/boolean_${value}.png`);
+        const sprite = Loader.getSprite(bitSpriteMap[this.value]);
         sprite.drawCachedCentered(parameters, x, y, diameter);
     }
 }
--- a/src/js/game/items/boolean_item.js
+++ b/src/js/game/items/boolean_item.js
@@ -4,6 +4,8 @@ import { types } from "../../savegame/serialization";
 import { BaseItem } from "../base_item";
 import { globalConfig } from "../../core/config";
 
+/** @typedef {0 | 1} Bit **/
+
 export class BooleanItem extends BaseItem {
     static getId() {
         return "boolean_item";
@@ -27,18 +29,18 @@ export class BooleanItem extends BaseItem {
     }
 
     /**
-     * @param {number} value
+     * @param {Bit} value
      */
     constructor(value) {
         super();
-        this.value = value ? 1 : 0;
+        this.value = value;
     }
 
     /**
-     * @param {BaseItem} other
+     * @param {BooleanItem} other
      */
     equalsImpl(other) {
-        return this.value === /** @type {BooleanItem} */ (other).value;
+        return this.value === other.value;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/js/game/items/boolean_item.js
```diff
@@ -48,15 +50,19 @@ export class BooleanItem extends BaseItem {
      * @param {DrawParameters} parameters
      */
     drawItemCenteredImpl(x, y, parameters, diameter = globalConfig.defaultItemDiameter) {
-        let sprite;
-        if (this.value) {
-            sprite = Loader.getSprite("sprites/wires/boolean_true.png");
-        } else {
-            sprite = Loader.getSprite("sprites/wires/boolean_false.png");
-        }
+        const value = Boolean(this.value).toString();
```
tobspr, line 53: I like this PR, except this part: I prefer manually specifying `true` and `false` for the string, so it's more clear which sprite is used.

src/js/game/systems/logic_gate.js
```diff
@@ -19,8 +18,7 @@ export class LogicGateSystem extends GameSystemWithFilter {
     }
 
     update() {
-        for (let i = 0; i < this.allEntities.length; ++i) {
-            const entity = this.allEntities[i];
+        this.allEntities.forEach(entity => {
```
tobspr, line 21: There is a reason why a for loop is used everywhere, and that is performance. `forEach` is simply too slow :)

src/js/game/systems/logic_gate.js
```diff
@@ -48,133 +46,84 @@ export class LogicGateSystem extends GameSystemWithFilter {
...
     compute_AND(parameters) {
         assert(parameters.length === 2, "bad parameter count for AND");
 
-        const param1 = parameters[0];
-        const param2 = parameters[1];
-        if (!param1 || !param2) {
-            // Not enough params
-            return BOOL_FALSE_SINGLETON;
-        }
+        const [param1, param2] = parameters;
```
tobspr, line 59: There's also a reason I don't use `[a, b] = array` in the whole codebase (with a few exceptions) since it's too slow too. (It gets polyfilled, and the polyfill is very slow)

src/js/game/items/boolean_item.js
```diff
@@ -48,15 +50,19 @@ export class BooleanItem extends BaseItem {
...
 
 export const BOOL_FALSE_SINGLETON = new BooleanItem(0);
 export const BOOL_TRUE_SINGLETON = new BooleanItem(1);
+
+/**
+ * @param {unknown} item
+ * @returns {item is BooleanItem}
+ **/
+export const isBooleanItem = item => {
+    return item instanceof BooleanItem;
```
tobspr, line 67: Don't use instanceof, that's too slow. Instead use `getItemType() === "boolean"`

src/js/game/systems/logic_gate.js
```diff
@@ -48,133 +46,84 @@ export class LogicGateSystem extends GameSystemWithFilter {
...
     }
 
     /**
      * @param {Array<BaseItem|null>} parameters
-     * @returns {BaseItem}
+     * @returns {BooleanItem}
      */
     compute_NOT(parameters) {
-        const item = parameters[0];
+        const [item] = parameters;
```
tobspr, line 73: Same here

src/js/game/systems/logic_gate.js
bjornstar: This is what the polyfill looks like:

```
  function Dw(t) {
      var e = t[0],
          r = t[1];
      return Object(hv.d)(e) && Object(hv.d)(r) && e.value && r.value ? hv.b : hv.a
  }
```

It is identical to the code I replaced.

src/js/game/items/boolean_item.js
bjornstar:
```
export const isBooleanItem = item => {
    return item instanceof BooleanItem;
};

const isItemType = item => {
    return item && item.getItemType() === "boolean";
};

import { COLOR_ITEM_SINGLETONS } from "./color_item";

window.benchmark1 = () => {
    console.time('instanceof');

    for (let i = 0; i < 100000; i += 1) {
        isBooleanItem(BOOL_FALSE_SINGLETON);
        isBooleanItem(COLOR_ITEM_SINGLETONS.red);
    }

    console.timeEnd('instanceof');
};

window.benchmark2 = () => {
    console.time('itemType');

    for (let i = 0; i < 100000; i += 1) {
        isItemType(BOOL_FALSE_SINGLETON);
        isItemType(COLOR_ITEM_SINGLETONS.red);
    }

    console.timeEnd('itemType');
};

for (let i = 0; i < 10; i += 1) {
    window.benchmark1();
    window.benchmark2();
}
```

We get results like this:
```
instanceof: 3.31494140625ms
itemType: 4.391357421875ms
instanceof: 7.876953125ms
itemType: 7.22509765625ms
instanceof: 0.050048828125ms
itemType: 0.0419921875ms
instanceof: 0.042236328125ms
itemType: 0.038818359375ms
instanceof: 0.0419921875ms
itemType: 0.038818359375ms
instanceof: 0.041015625ms
itemType: 0.0400390625ms
instanceof: 0.0419921875ms
itemType: 0.0458984375ms
instanceof: 0.0478515625ms
itemType: 0.044921875ms
instanceof: 0.05712890625ms
itemType: 0.046142578125ms
instanceof: 0.048095703125ms
itemType: 0.046142578125ms
```
 The difference is negligible, especially when v8 optimizes the function.

The advantage of the `instanceof` check is that we can pass anything in there, while if we rely on the class method, we have to be sure that we do a null check first.

tobspr: Interesting! is it also the case on firefox and older browsers?

bjornstar: On Firefox where the game greets you with this message:

shapez-chrome

```
instanceof: 2ms - timer ended
itemType: 3ms - timer ended
instanceof: 3ms - timer ended
itemType: 3ms - timer ended
instanceof: 2ms - timer ended
itemType: 0ms - timer ended
instanceof: 2ms - timer ended
itemType: 1ms - timer ended
instanceof: 1ms - timer ended
itemType: 0ms - timer ended
instanceof: 2ms - timer ended
itemType: 0ms - timer ended
instanceof: 2ms - timer ended
itemType: 0ms - timer ended
instanceof: 3ms - timer ended
itemType: 1ms - timer ended
instanceof: 2ms - timer ended
itemType: 0ms - timer ended
instanceof: 2ms - timer ended
itemType: 1ms - timer ended
```

Issue #594, opened 2020-08-25
KeyUp event ignored while context menu is opened


bcmpinc:
**How to reproduce**
1. Select belt
2. Shift right click somewhere
3. See the context menu popup
4. Release the shift key
5. Close the context menu
6. Note that belt is no longer selected
7. Select belt
8. Note that belt is shown as a green dot
9. You can now place belts in planner mode while no longer having to hold shift.
10. Note that when scrolling the map it also moves faster as if shift is still being pressed.

I suspect that the KeyUp event for shift is ignored while the context menu is open.

Issue #584, opened 2020-08-21
FEATURE REQUEST: Sticky/Toggle Planning Mode


iansebryk:
can we please have the ability to toggle (make sticky) planning mode?  having to hold shift all the time is tedious.  thank you.

PR #595, opened 2020-08-25
Releasing SHIFT no longer cancels the belt planner

bcmpinc:
This (sort of) implements #584 by only needing to hold shift while starting to use the belt planner. To cancel the belt planner, use right-click. It is implemented by setting `this.currentDirectionLockDragging` while the player is dragging in direction lock/planning mode.

As this causes SHIFT and right click to be more likely to be used together, it would make bug #594 more noticeable.

In `abortDragging()` I changed `this.currentlyDragging = true;` to `this.currentlyDragging = false;`, as I believed this was a programming error. Please check.


tobspr:
Hm, this should at least be a setting

tobspr:
Closing until changes are made :)

PR #598, opened 2020-08-27
Update base-cz.yaml

Remogeus:
translated terms for wires, description for a marker, categories in settings, quad painting title, reward text for a painter and confirmation of mass cut while not enough blueprints are available


tobspr:
Ty!

PR #603, opened 2020-08-28
Changed wrong naming convention in README.md

TehThanos:
The README said to use camel case names for new components while it should be pascal case.


tobspr:
Ty!

PR #606, opened 2020-08-28
for prep to upgrade

jimmykarns:


jimmykarns:
actually might want to ignore this, may require a thorough check

tobspr:
This branch is empty ...

PR #607, opened 2020-08-28
"Logic pain fix" - Fixing quad painter pain...

dengr1065:
First of all, this change makes 4-painter ignore or trash paint if there are missing quadrants. Also, it allows players to use wired pins for disabling and filtering inputs (paint is trashed if unmatched, too). This is done by adding `ProcessingRequirement` component and additional checks in quad painter processor. Note that ProcessingRequirement can be also useful for things like blocking trash and other processors depending on other factors. The quad painter trashes unmatched colors just to make sure inputs are never blocked and/or have weird results.

This change should be tested extensively in alpha, because I checked just basic shapes such as `RuRuRuRu` and `--RuRuRu:RuRu----`.


src/js/game/components/processing_requirement.js
```diff
@@ -0,0 +1,90 @@
...
+
+        // Type of the processor
+        this.type = processorType;
+    }
+
+    /**
+     * Checks whether it's possible to process something
+     * @param {Entity} entity
+     */
+    canProcess(entity) {
```
tobspr, line 36: Hm, I don't like that there is so much logic in this method. Could it be moved into a game system?

tobspr (changes requested):

tobspr:
I think, given that it belongs to the item processor, why not add a new `enumItemProcessorRequirement` which can be set on the `ItemProcessor` component as a `processingRequirement` ?

dengr1065:
> I think, given that it belongs to the item processor, why not add a new `enumItemProcessorRequirement` which can be set on the `ItemProcessor` component as a `processingRequirement` ?

Thanks, that sounds good, how could I forget....

tobspr:
The item processor system could then have a `canStartProcessing` method (Idk about the name) which is called by the processor. This could also be used for the item filter I think.

dengr1065:
While refactoring, i found some little bug, thanks :smile:

dengr1065:
working now - pushing!
<table>
<tr>
<td>

</td>
<td>

</td>
</tr>
</table>

dengr1065:
I wonder why does linting fail remotely but `prettier-all` works just nice...

src/js/game/components/processing_requirement.js
dengr1065: moved to itemprocessor system as separate method

tobspr:
Thanks! I'll fix the linter issues

PR #609, opened 2020-08-30
Update base-de.yaml

DerFeldspatz:
The store description got changed back to an older version in the last PRs. These included various mistakes, so I fixed them. There is also some rearrangement in the shortcut section, so it matches the english base again. Also changed the `noShapesProduced` translation so the message makes at least some sense for all 3 tabs in statistics window. (You could implement different messages for them, so they make sense). Feel free to review and discuss!


tobspr:
Danke!

PR #610, opened 2020-08-30
Update base-uk.yaml

Prosta4okua:


tobspr:
Please don't commit your config.local.js :)

Prosta4okua:
> Please don't commit your config.local.js :)
@tobspr
I didn't edit that file. What should I do?

PR #615, opened 2020-08-31
Updated base-da.yaml

jakob753951:
Made the names of buildings and levels more internally consistent, Fixed some grammar, and re-did some awkward wordings.


tobspr:
Thanks!

PR #617, opened 2020-08-31
Update base-de.yaml

DerFeldspatz:
Typos, Grammar, concistency of punctuation and `drück(e)`. Some phrasing and a suggestion for the translation of new volume feature. @Yoshie2000 please check if it's ok :) 
Feedback and reviews appreciated.

### DIFF
--- a/translations/base-de.yaml
+++ b/translations/base-de.yaml
@@ -204,7 +204,7 @@ dialogs:
 
     editKeybinding:
         title: Tastenbelegung ändern
-        desc: Drück die (Maus-)Taste, die du belegen möchtest, oder ESC um abzubrechen.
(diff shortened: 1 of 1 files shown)


Yoshie2000:
This is fine with me, I like this translation a lot better than my original one.
I'm not quite sure if we should replace `drücke` with `drück`. Technically both is correct, I personally prefer `drücke`, but I don't mind it being `drück` either...

DerFeldspatz:
I had to start somewhere 😅

DerFeldspatz:
It's changed to `Drücke` now consisently.

tobspr:
Vielen Dank!

Issue #620, opened 2020-08-31
[Feature Request] Modify opacity to allow placing buildings under Placement Hints


mvb0005:
Currently, it is quite difficult to place buildings under the placement hints hud. My proposed change would be to reduce the opacity on hover.

PR #621, opened 2020-08-31
Change opacity on hover

mvb0005:
Code change for Issue #620 

I modified the opacity of the hud, but the picture was still distracting, so I also removed the picture all together on hover.
These values are not set in stone, and I am open to suggestions.

image


tobspr:
oh well, thats pretty smart!

PR #629, opened 2020-09-02
Improve French translation

LeopoldTal:
Commits are ordered from most straightforward to most controversial, so you can drop some if you don't like a change.

* Obvious typo etc. fixes
* Rename wire to "câble" (shorter, closer to literal). @PascalGrosse @martypiton does that sound good to you?
* Typographical pedantry: prefer sentence case over title case, punctuation, non-breaking spaces
* Heavier rephrasing for consistent, idiomatic, and clear translation — bikeshedding goes here

I have **not** synced translations, just tried to improve whatever exists now including obsolete keys.


translations/base-fr.yaml
```diff
@@ -289,13 +289,13 @@ ingame:
...
-        lockBeltDirection: Utiliser le plannificateur de convoyeurs
-        plannerSwitchSide: Échanger la direction du plannificateur
+        lockBeltDirection: Utiliser le planificateur de convoyeurs
+        plannerSwitchSide: Inverser la direction du planificateur
         cutSelection: Couper
         copySelection: Copier
         clearSelection: Effacer la sélection
         pipette: Pipette
-        switchLayers: Échanger les calques
+        switchLayers: Changer de calque
```
LeopoldTal, line 298: "Layer" here would be better translated as "couche" but that evokes diapers. Rethink after wires update?

PascalGrosse: Well, I chose "calque" because it's the word used as translation of "layer" in most software using this functionality. The literal translation isn't always the best, especially in the case of "couche"/"diapers" :-)

BTW, I think Échanger is better than Changer, because to switch != to change in english. The semantic isn't the same. And I'm not even talking of "Changer de couche"... "Basculer le calque" would be even better I think.

PascalGrosse: But I'm probably a little bit too pedantic (and maybe even wrong), your translations are very good !

LeopoldTal: Re: "layer": I'm aware of "calque" used in image editing software, but here the metaphor is more of literal layers of terrain, with underground wires. "Niveau" would work if it wasn't [already used for game levels](https://www.giantitp.com/comics/oots0012.html). Had a look at what SimCity uses for the water layer and it doesn't talk about layers at all, just "vue"/"affichage"/"carte".

Re: "switch": To me "échanger" suggests it affects the game state (somehow swaps the layers), not just the interface. "Basculer" is very good, thanks. But it will need to be changed if there are ever more than 2 layers.

PascalGrosse: You're right. At the time, the intended game semantic wasn't even properly specified (is it now ?). I quite sure translations would need to be adjusted further in the development line.

Naheulf:
Please wait for https://github.com/LeopoldTal/shapez.io/pull/1

translations/base-fr.yaml
```diff
@@ -49,29 +49,29 @@ steamPage:
...
+            [*] Prochainement : Câbles et énergie ! Prévu pour (environ) fin juillet 2020.
+            [*] Prochainement : Plus de niveaux
             [*] Aidez moi à continuer de développer shapez.io ❤️
         [/list]
 
-        [b]Mises à jour futures[/b]
+        [b]Mises à jour à venir[/b]
 
-        Je fais souvent des mises à jour et essaye d'en sortir une par semaine!
+        Je fais souvent des mises à jour et essaye d'en sortir une par semaine !
```
martypiton, line 59:
```suggestion
        Je fais souvent des mises à jour et j'essaye d'en sortir une par semaine !
```
You can't skip the subject here, you must put it a second time otherwise it's not french :)

LeopoldTal:
> Please wait for [LeopoldTal#1](https://github.com/LeopoldTal/shapez.io/pull/1)

Merged!

translations/base-fr.yaml
LeopoldTal: Done, but with Naheulf's fancy typesetter's apostrophe to be consistent.

(Also, why say it's "not French"? There are many examples of native French speakers using such constructions. Can't you just argue it's better style?)

tobspr:
Uh, I think it conflicted with the other PR I just merged, could you have a look? Thanks in advance!

translations/base-fr.yaml
```diff
@@ -728,75 +729,75 @@ settings:
...
             description: >-
-                Desactiver la grille de placement peut aider les performances. Ça rend aussi le jeu encore plus uni!
+                Désactiver la grille de placement peut améliorer les performances. Ça rend aussi l’apparence plus unie !
 
         clearCursorOnDeleteWhilePlacing:
-            title: Effacer le curseur avec clic droit
+            title: Vider le curseur avec le clic droit
             description: >-
-                Activé par défaut, efface le curseur lorsque vous faites un clic droit en ayant un bâtiment selectioné pour la constructio. Si desactivé, vous pouvez detruire les bâtiments en faisant un clic droit tout en placant un bâtiment.
+                Activé par défaut, remet à zéro le curseur lorsque vous faites un clic droit en ayant un bâtiment selectionné pour la construction. Si désactivé, vous pouvez détruire les bâtiments avec un clic droit puis continuer de placer un bâtiment.
```
LeopoldTal, line 785: Not very happy with this, @Naheulf any better ideas?

Naheulf: I think can be explicit to the current behavior but is pretty long:
> Activé par défaut. Désélectionne le bâtiment choisi pour la construction lorsque vous faites un clic droit sur un bâtiment existant. Si désactivé, vous pouvez détruire des bâtiments avec un clic droit puis continuer de placer le bâtiment sélectionné.

tobspr:
Thanks a lot!

PR #649, opened 2020-09-12
Add a setting to disable picking the miner when on a resource patch

Exund:
I found the automatic miner picker when using the pipette on a resource patch annoying but I can't disable it. I totally see why it can be nice to have but making it a setting doesn't take much.


tobspr:
Thanks!

PR #660, opened 2020-09-16
Update base-ind.yaml

FahrulID:
Change a lil bit, might change more later

### DIFF
--- a/translations/base-ind.yaml
+++ b/translations/base-ind.yaml
@@ -22,10 +22,10 @@
 ---
 steamPage:
     # This is the short text appearing on the steam page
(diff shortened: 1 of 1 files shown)


translations/base-ind.yaml
```diff
@@ -142,7 +142,7 @@ mainMenu:
     play: Mulai Permainan
     continue: Lanjutkan Permainan
     newGame: Permainan Baru
-    changelog: Ganti Data Log
+    changelog: Catatan Perubahan
```
DerFeldspatz, line 145: You added an unnecessary space after the translation. These are called `trailing spaces` and cause the build to produce error messages. Please remove it and check for them in the future

tobspr:
Thanks! I'll fix the space

PR #665, opened 2020-09-17
Better Entity Debugger, attempt 2

dengr1065:
Right now, it will be enabled as earlier - only in dev mode with a flag in config. In short: this displays component properties as a tree, instead of serialized JSON.
<details><summary>Screenshot</summary>Entity Debugger on Hub</details>


tobspr:
Awesome, thanks!

PR #666, opened 2020-09-17
Building variants refactor

Exund:
Delegate variants to classes instead of relying on switch-cases

Allows for easier variant refactoring and addition

Also changed building codes to string to allow more flexibility


tobspr:
Hey, first of all this is a super cool PR! 

Refactoring the variants has on been on my todo list for a long time.

I like that you made them classes - but why not do the next step and instead of having a variant use the meta building itself?

I.e. you don't pass the variant anywhere, instead just the meta building. What do you think?

tobspr:
Closing for now, but its still in my mind and linked on trello for reference

PR #669, opened 2020-09-19
Another small French translation fix

LeopoldTal:
"Créé le" must be followed by an absolute date, but this is always interpolated with a relative date, e.g. "il y a 42 jours", so the article is redundant.

I've had a look at all other French translations containing variables and found no other problems.


tobspr:
Thank you!

PR #670, opened 2020-09-19
Fix build errors

LeopoldTal:
* YAML lint: Remove trailing spaces introduced in 7d6af359a192bc6aa27dcaaa366bd58c45266a89
* ~~ESLint: Buffer stats now contain  `backlogKeys` and `backlogSize` instead of `backlog`, but the logger still tries to log `backlog`. Lint fails and it logs "undefined". Logging both of the new stats instead.~~ Never mind, you did that in 1cd52f6dbdc94a28684576296e3c3091d6e75b57

### DIFF
--- a/translations/base-en.yaml
+++ b/translations/base-en.yaml
@@ -191,7 +191,7 @@ dialogs:
     confirmSavegameDelete:
         title: Confirm deletion
         text: >-
-            Are you sure you want to delete the following game?<br><br>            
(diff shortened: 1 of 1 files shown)


LeopoldTal:
Blech, eslint also broken. I'll propose a fix but I'm not actually sure how you want the new buffer stats to be logged.

tobspr:
Ty!

PR #671, opened 2020-09-19
Refactor entity cloning

Exund:
Removed `duplicateWithoutContents` from all components and added an optional `copyAdditionalStateTo(otherComponent)` method to copy addition data (lever state, constant signal value, etc.)

### DIFF
--- a/src/js/game/entity.js
+++ b/src/js/game/entity.js
@@ -11,6 +11,7 @@ import { EntityComponentStorage } from "./entity_components";
 import { Loader } from "../core/loader";
 import { drawRotatedSprite } from "../core/draw_utils";
 import { gComponentRegistry } from "../core/global_registries";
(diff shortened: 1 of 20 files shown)


src/js/game/entity.js
```diff
@@ -82,14 +83,25 @@ export class Entity extends BasicSerializableObject {
...
      */
-    duplicateWithoutContents() {
-        const clone = new Entity(this.root);
+    clone() {
+        const staticComp = this.components.StaticMapEntity;
+        const buildingData = getBuildingDataFromCode(staticComp.code);
+
+        const clone = buildingData.metaInstance.createEntity({
+            root: this.root,
+            origin: staticComp.origin,
```
tobspr, line 94: You need to clone the vector here too, since otherwise changing it would change the original vector

tobspr (changes requested):

PR #672, opened 2020-09-19
Show Display and Lever values on map view (+1 commit)

### LINKED ISSUE #657: Feature request: display status visible from map view
Currently the display buildings all show as the same color once zoomed out to map view, regardless of whether they are on or off. 
It would be pretty cool if the color of the map tile changed appropriately depending on whether the display was on or off.  This would be great for making large blinkenlights displays and viewing while zoomed way out.

For boolean true vs false (or not connected) it could just toggle between two lighter/darker colors.
If the display signal is a color, then showing the actual corresponding color would be great too.
If the display signal is a shape type, then they would probably be too tiny to discern from map view, in that case, just a single additional color for active vs inactive could still be useful.

dengr1065:
changes:
- Map view for displays and levers
- Remove unused method and property

made from github cli, too lazy
// Fixes #657

### DIFF
--- a/src/js/game/systems/display.js
+++ b/src/js/game/systems/display.js
@@ -1,7 +1,7 @@
 import { globalConfig } from "../../core/config";
 import { Loader } from "../../core/loader";
 import { BaseItem } from "../base_item";
-import { enumColors } from "../colors";
+import { enumColors, enumColorsToHexCode } from "../colors";
 import { DisplayComponent } from "../components/display";
 import { GameSystemWithFilter } from "../game_system_with_filter";
 import { isTrueItem } from "../items/boolean_item";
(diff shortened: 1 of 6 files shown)


dengr1065:
Fixes #657

src/js/game/systems/display.js
```diff
@@ -94,4 +94,50 @@ export class DisplaySystem extends GameSystemWithFilter {
...
+            }
+
+            const value = this.getDisplayItem(network.currentValue);
+
+            if (!value) {
+                continue;
+            }
+
+            const origin = entity.components.StaticMapEntity.origin;
+            if (value instanceof ColorItem) {
```
tobspr, line 125: use `value.getItemType() === "color"`

src/js/game/systems/display.js
```diff
@@ -94,4 +94,50 @@ export class DisplaySystem extends GameSystemWithFilter {
...
+        for (let i = 0; i < contents.length; ++i) {
+            const entity = contents[i];
+            if (!entity || !entity.components.Display) {
+                continue;
+            }
+
+            const pinsComp = entity.components.WiredPins;
+            const network = pinsComp.slots[0].linkedNetwork;
+
+            if (!network || !network.currentValue) {
```
tobspr, line 114: You can (and should) use `!network || !network.hasValue())`

src/js/game/systems/display.js
```diff
@@ -94,4 +94,50 @@ export class DisplaySystem extends GameSystemWithFilter {
...
+            const origin = entity.components.StaticMapEntity.origin;
+            if (value instanceof ColorItem) {
+                this.displaySprites[value.color].drawCachedCentered(
+                    parameters,
+                    (origin.x + 0.5) * globalConfig.tileSize,
+                    (origin.y + 0.5) * globalConfig.tileSize,
+                    globalConfig.tileSize
+                );
+
+                parameters.context.fillStyle = enumColorsToHexCode[value.color];
```
tobspr, line 133: Why are you drawing the display sprite here as well as fillRect?

tobspr (changes requested):

tobspr:
Nice PR! However I think this might be too slow. `fillRect` for example is super slow.

I'd like to have a solution where additional to the cached chunk there is a secondary cached layer which is regenerated automatically once the value changes. If there are no wires components in that chunk, it can also entirely be skipped.

What do you think?

src/js/game/systems/display.js
dengr1065: instanceof makes more sense from oop point, and using `.getItemType` breaks lint -- you can make a webpack define that converts `instanceof *Item` to these calls, though.

src/js/game/systems/display.js
dengr1065: lol, weird copy paste i guess

tobspr:
Still, its not performant :p

dengr1065:
Oh, yeah, about `fillRect`: should do some tests in-game in these cases:
 - 5-10 displays on in 4 chunks (20-40 displays total)
 - 1 chunk with lots of displays

These should be most used ones (marking and screens). I can't understand how `fillRect` can be slower than caching all textures and repainting them each time: even with fast (aliased) zooming, there are more calculation, whereas `fillRect` should compute just bounds, and fill them. `uncolored`/`0` displays are not rendered at all, which is also a popular usage for screens (16+ displays/chk)

Maybe let's try it in beta? It should be performant enough, even if `fillRect`, for some reason, is slower (that may depend on OpenGL implementation too).

P.S: I think, there's a way to merge all displays in a chunk to some "all" matrix, so there is just 1 object per chunk. But as I know, drawing paths is even slower.

tobspr:
I'll try to figure sth, probably not before the next update tho

tobspr:
Closing for now since I want to do it differently

PR #673, opened 2020-09-19
Fix camera pan

Exund:
You could still move when the shape viewer was opened
You were also able to leave the browser viewport and still have the camera pan

### DIFF
--- a/src/js/game/camera.js
+++ b/src/js/game/camera.js
@@ -862,7 +862,11 @@ export class Camera extends BasicSerializableObject {
      * @param {number} dt
      */
     internalUpdateMousePanning(now, dt) {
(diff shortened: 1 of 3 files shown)


src/js/game/camera.js
```diff
@@ -862,7 +862,7 @@ export class Camera extends BasicSerializableObject {
      * @param {number} dt
      */
     internalUpdateMousePanning(now, dt) {
-        if (!this.root.app.settings.getAllSettings().enableMousePan) {
+        if (!this.root.app.settings.getAllSettings().enableMousePan || !this.root.app.focused) {
```
tobspr, line 865: I'd prefer a separate `if` but its not a huge issue

tobspr (changes requested):

PR #674, opened 2020-09-20
fix build for mac os

smartmanru:
without it on the latest version of OS X they were not going to what


LeopoldTal:
High-quality sounds are causing problems for some users on OSX, but that's no reason to remove them for everyone!

tobspr:
Yup, that's not a good fix

PR #675, opened 2020-09-20
Fix 180° rotater not renamed everywhere

Exund:
Fixed atlas and buildingSpeed


tobspr:
Ah, thanks! Missed that :P

tobspr:
You don't need to save the atlas.json, since it is automatically generated. Could you remove it from the PR ? :)

tobspr:
Ty!

PR #676, opened 2020-09-20
Fix size overflow when building sounds

LeopoldTal:
# Issue

Building the standalone with V8 v8 (Node v14) fails at step `sounds.musicHQ`, either by running out of memory or with the error "Invalid string length". Building with V8 v7 (Node v12) succeeds.

# Cause

This is because `gulp-cache` builds a 994-MB string when [stringifying file contents](https://github.com/jgable/gulp-cache/blob/master/src/task-proxy.js#L262) to cache them. V8 v8 [limits string length to around 537 MB](https://github.com/v8/v8/blob/master/src/objects/string.h#L384), and thus cannot represent this string. (V8 v7 allows around 1074 MB, which is why the build passes on Node v12.)

But [`theme-full.mp3`](https://github.com/tobspr/shapez.io/blob/master/res_raw/sounds/music/theme-full.mp3) is only 79 MB: how did we get 1250% overhead?

Unlike plaintext files, binary files are read as buffers, but [by default](https://github.com/jgable/gulp-cache/blob/master/src/index.js#L46) `gulp-cache` stringifies them naively, producing the extremely inefficient representation:

````
[
  {
    "cwd": "/Users/tobspr/shapez.io/gulp",
    "base": "/Users/tobspr/shapez.io/res_raw/sounds/music",
    "contents": {
      "type": "Buffer",
      "data": [
        73,
        68,
        51,
        4,
        0,
        0,
…etc.
````

# Fix

Fortunately, `gulp-cache` [can read base64-encoded cache files](https://github.com/jgable/gulp-cache/blob/master/src/index.js#L26).

Instead of using the default file transform function, **pass a `value` option to base64-encode the file contents**. This results in only 33% overhead on cache file size (106 MB for the largest file).

This has multiple benefits:

- Fixes the build failure
- Requires less memory (from 6 GB down to < 1 GB on my machine)
- When cache files are found, the `sounds.musicHQ` is much faster (from ~30 s down to ~4 s on my machine)
- Smaller cache files on disk

Also note that this only affect cache writes, so it won't invalidate an existing cache.

# Testing

Tested on OS X. Please make sure to test:

- on Windows and Linux
- at least on Node v14, ideally also v12
- both web build and standalone build

### DIFF
--- a/gulp/sounds.js
+++ b/gulp/sounds.js
@@ -16,6 +16,12 @@ function gulptasksSounds($, gulp, buildFolder) {
         cacheDirName: "shapezio-precompiled-sounds",
     });
 
+    function getFileCacheValue(file) {
+        const { _isVinyl, base, cwd, contents, history, stat, path } = file;
+        const encodedContents = Buffer.from(contents).toString('base64');
+        return { _isVinyl, base, cwd, contents: encodedContents, history, stat, path };
+    }
+
     // Encodes the game music
     gulp.task("sounds.music", () => {
         return gulp
@@ -34,6 +40,7 @@ function gulptasksSounds($, gulp, buildFolder) {
                     {
                         name: "music",
                         fileCache,
+                        value: getFileCacheValue,
                     }
                 )
             )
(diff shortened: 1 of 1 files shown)


tobspr:
Very well made/explained PR, thank you! I am using node v12  which is why it wasn't an issue for me, but I guess it should also speed up the build by a lot.

I'll also test it on windows and on the final builds :)

PR #679, opened 2020-09-21
Added greek translation

TCLamnidis:
I tried to stay as close to the original English as possible. It would be nice to also see the translated version in the game to determine how/where these appear in the game itself. That would allow me to fix any tenses etc that may be hard to get right without all the context. 

This should still be mostly good!

Thanks for making this great game!


tobspr:
Ty!

PR #680, opened 2020-09-21
FIX screenshot. Use same scale for Atlas==zoomLvl

przemko07:
Scale 1 is never gonnna be saved as screenshot. for most image exports 0.5 is used. chunkScale is determining that.


tobspr:
Awesome, thanks! I Totally missed that

PR #683, opened 2020-09-23
Improved and updated spanish translation

Galo223344:
Fixed some unnecessary extra spaces, translated pipette, improved trash translation (It was worded in a way as if it destroyed "entrances") also translated "advanced" wich was untranslated for some reason.


Galo223344:
It seems like the check failed because of base-el 🤔 (This is base-es)

tobspr:
Thanks!

PR #691, opened 2020-09-26
Update base-pt-BR.yaml

CFWarfare:
Fixed some phrases that wouldn't be used in the language regularly (like 'Código Aberto')


tobspr:
Thanks!

PR #698, opened 2020-09-28
Make numbers display to set amount of significant figures

EmeraldBlock:
This modifies the formatBigNumber function so that it outputs numbers to a set amount of significant figures (default of 3). It also has a new optional argument for the number of significant figures.

The hub delivered text size is also made smaller (16px starting from 1000 instead of 10000) to accommodate the additional digit.

image
image

This is my first PR ever, and I tested it through an unorthodox method, so there may be some issues.


EmeraldBlock:
The number's length sort of "pulses" whenever it hits a rounder number, it might be better to display trailing zeroes, but not sure

tobspr:
Hm I'll need to test it first, on my todo list

EmeraldBlock:
oops, committed something to the wrong branch

tobspr:
Not planning to merge right now, thus closing - I don't see the benefit currently

PR #700, opened 2020-09-28
Ability to edit Constant Signals

Git-Nivrak:


tobspr:
Can you have a look at the linter issues first? :)

daanbreur:
May I take a look at this??

Git-Nivrak:
Are the YAML lint errors an issue too?

daanbreur:
No.

tobspr:
Uh so, I don't really like storing a `clicked` flag on the constant signal.

Instead, why not call `this.root.systemMgr.systems.constantSignal.triggerSignalEdit()`? 

Actually the clean way (Which I failed my self to do and I can fix too :P) would be to have the whole editing of the constant signal within a HUD component and have the HUD component listen on that.

Let me know if you want to do it or I should do it

Git-Nivrak:
^ Like so?

tobspr:
Similar, since the creation and editing of constant signals is the same, I think the code should be shared

Git-Nivrak:
Not exactly the same, You don't want it to destroy the object when they press cancel once it is already placed

Git-Nivrak:
Maybe I'll add an optional argument for if it is editing or creating

daanbreur:
@tobspr is right. The code should be shared

Git-Nivrak:
Here

tobspr:
I recommend to install eslint and prettier, there are a lot of linter issues again ! :) :P

Git-Nivrak:
I have them downloaded, I thought the lint YAML wasn't an Issue?

tobspr:
No, not YAML, but seee the github comments :)

tobspr:
Looks good now! Probably won't make it into the wires update though, depends on how much time I have left

Git-Nivrak:
Soon™

tobspr:
Sooner™

Git-Nivrak:
Soonest™

darkwater4213:
Yay! I like this idea too.

tobspr:
Done in 1.2.2

PR #704, opened 2020-09-28
Make zoom in and zoom out equal

EmeraldBlock:
Makes zooming out zoom an amount equal to zooming in. This changes both mouse zoom and +/- zoom. All zoom that uses a smooth transition is now exact. (Rapidly zooming in and out will still not guarantee equal zoom, since the transitions are canceled.)

zoomfix2

Also tested through an unorthodox method


tobspr:
Awesome, thanks!

PR #712, opened 2020-09-29
Add some Japanese text

Starwort:
I'm still learning so the long sentences are too complex for me, but I did some shorter keys. I hope it helps :)


tobspr:
Thank you! ❤️

PR #713, opened 2020-09-29
Started translation of intro and related sections

Eble6:
Started on the translation of intro and related sections like "Standalone advantages


tobspr:
Thank you! ❤️

PR #715, opened 2020-09-29
Update to base-pl.yaml

SquirrelWarrior13247:
Translated some things, mostly at the beginning and 360 - 500.


tobspr:
Thank you! It seems somebody was faster though .. :P Could you have a look? :)

PR #716, opened 2020-09-29
Update base-fr.yaml

Malak044:


LeopoldTal:
Ah, sorry to duplicate but a bunch of us on the Discord have made huge changes (#718). Want to leave comments on there instead?

tobspr:
See comment above! :)

Malak044:
yeah well it's, fine.  Yours seems better as I went for some literal traduction and felt a bit awkward about those.
Let's close this.

PR #717, opened 2020-09-29
Updated and improved some zh-CN translations.

AAAshuibiao:
Updated and improved some zh-CN translations. The translations are still incomplete and needed to be worked on in the future.


tobspr:
Thank you! ❤️

PR #720, opened 2020-09-30
Update base-cat.yaml

JSansoP:
Translated all new options and items/features. Left steampage and free version translations untranslated because they get reseted every time there is a new version. Tobias, if you want them translated contact me on discord ItzSanso#6913 and i will do it whenver i can.


tobspr:
Thank you! The steam page is now finalized for this version, and yeah I clear it once its outdated, would be nice if you could have a look now! :)

PR #727, opened 2020-09-30
Dockerfile use for Node12

endofzero:
I've been using this dockerfile successfully to build and run Shapez.io without needing to install and maintain node.

Building is straight forward: "docker build -t shapezio ."

After building you and start and stop the image whenever you need to.

Running is just as straight forward: "docker run --name="shapezio" --rm -p 3005:3005 shapezio"

To stop the instance: "docker stop shapezio"

Due to --rm the instance should be deleted once stopped.

If you want to stop the messages about not being able to open the browser, you can edit line 155 of gulp/gulpfile.js, and add
"open: false," 
It shouldn't affect anything having it still there, but altering it may mess up future pull requests so I would suggest leaving it as is and just ignore the warning.

function serve({ standalone }) {
    browserSync.init({
        server: buildFolder,
        port: 3005,
        ghostMode: {
            clicks: false,
            scroll: false,
            location: false,
            forms: false,
        },
        logLevel: "info",
        logPrefix: "BS",
        online: false,
        **open: false,**
        xip: false,
        notify: false,
        reloadDebounce: 100,
        reloadOnRestart: true,
        watchEvents: ["add", "change"],
    });

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,15 +15,10 @@ RUN apt-get update \
     ffmpeg \
     && rm -rf /var/lib/apt/lists/* 
 
-RUN apt-get update \
-    && apt-get install -y --no-install-recommends \
-    git-lfs
-
 FROM base as shape_base
 
 WORKDIR /shapez.io
 
-RUN git-lfs pull
 RUN yarn
 
 WORKDIR /shapez.io/gulp
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,6 @@ RUN apt-get update \
     && apt-get upgrade -y \
     && apt-get dist-upgrade -y \
     && apt-get install -y --no-install-recommends \
-    git \
     ffmpeg \
     && rm -rf /var/lib/apt/lists/* \
     && curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | bash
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+    git-lfs
+
+FROM base as shape_base
+
+WORKDIR /shapez.io
+
+RUN git-lfs pull
```
tobspr, line 27: I just recently got rid of git-lfs, so I think you can remove that :)

tobspr (changes requested):

tobspr:
Looks good except for a minor change! :)

Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+EXPOSE 3001
+
+RUN apt-get update \
+    && apt-get update \
+    && apt-get upgrade -y \
+    && apt-get dist-upgrade -y \
+    && apt-get install -y --no-install-recommends \
+    ffmpeg \
+    && rm -rf /var/lib/apt/lists/* \
+    && curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | bash
```
endofzero, line 17:
```suggestion
    && rm -rf /var/lib/apt/lists/* 
```

tobspr:
There's still a git-lfs pull command inside :)

tobspr:
Nice, thank you!

PR #731, opened 2020-10-01
Don't try to launch git lfs from gulpfile.js

dengr1065:
Cloning repo is nearly impossible while having Git LFS hooks installed, and most
Linux distros also don't have Git LFS installed by default. When running Gulp with
Git LFS without hooks, they'll get installed, thus breaking checkouts again.

Made this PR from web interface because my clone is broken again... :slightly_frowning_face:


tobspr:
Ah, thx! :)

PR #732, opened 2020-10-01
Update base-tr.yalm

TcePrepK:
I updated some of the translations and there was items not translated I translated them


tobspr:
Thank you!

PR #734, opened 2020-10-01
Update base-jp.yaml

Nucleareal:
The changelog is complex but it contains reorder keys by Original(base_en)
and some new keys for wires update


Nucleareal:
it has a conflict and closed sorry

PR #736, opened 2020-10-01
Update base-ja.yaml

Nucleareal:
Resolved conflicts

The changelog is complex but it contains keys reorder based on Original(base_en)
and some new keys for wires update


tobspr:
Awesome, thank you!

PR #738, opened 2020-10-01
Small improvements to French translation

LeopoldTal:
- restore missing translations
- rename "moulin" to "hélice"
- use "supprimer" rather than "effacer" when appropriate
- abbreviate "maximal" to fit in available space


tobspr:
Thank you!

PR #739, opened 2020-10-01
Specify path to config file in build instructions

### LINKED ISSUE #730: unclear instruction in README
I've noticed some things that are not clear in the project's readme (coming from someone who is not a JS programmer so JS-related things aren't obvious to me):

* some distros do not include `git-lfs` by default, you may want to clarify that it's required alongside `ffmpeg`, `node` and `yarn`
* it's not clear where `config.js` is located (I see `tsconfig.json` and `jsconfig.json`, are those related?), how to edit it and if I have to do that before or after running yarn in one or both folders.

LeopoldTal:
Closes #730.


tobspr:
Thank you!

PR #740, opened 2020-10-01
Add more polish translations & polish new polish translations

mini-bomba:
Since the translations were finally synchronized, I decided to add new polish translations.
Also changed some current translations a bit (polished them)


tobspr:
Awesome, thank you!

PR #743, opened 2020-10-01
Update base-cz.yaml

RevosCZ:
I made some corrections and added some further translations, later also repaired some mistakes in my own translations.


tobspr:
Awesome, thank you!

PR #744, opened 2020-10-02
Fix YAML errors with cz and pt-PT translations

3ddyy:
The game builds when this is broken, but opening it (at least the standalone build) gets you stuck on the loading screen


tobspr:
Thanks!

PR #747, opened 2020-10-03
Update docString for nextIntRangeInclusive to note un-usual distribution

macdja38:
I was going through the shape generation code and I found this random number generation code super strange due to the non uniform distribution of numbers creates, wanted to add a comment to make it more obvious.

If this isn't intended let me know and I'll change the PR to fix the RNG.

### DIFF
--- a/src/js/core/rng.js
+++ b/src/js/core/rng.js
@@ -108,7 +108,12 @@ export class RandomNumberGenerator {
         assert(max > min, "rng: max <= min");
         return Math.floor(this.next() * (max - min) + min);
     }
+
     /**
+     * Generates a number between min and max, inclusive.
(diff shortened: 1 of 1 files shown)


EmeraldBlock:
It appears that `nextIntRangeInclusive` is *only* ever used when determining the color palette for a freeplay shape. It may just be a good idea to just change
```js
const index = rng.nextIntRangeInclusive(0, colorWheel.length - 3);
```
to
```js
const index = rng.nextIntRange(0, colorWheel.length - 2);
```
and remove the `nextIntRangeInclusive` function.

tobspr:
What @EmeraldBlock  said

tobspr:
Closing for now, added as trello card

PR #752, opened 2020-10-03
Simplify build process

dengr1065:
I hate modules such as `trim`, so I migrated everything to use
`String.prototype.trim` instead. I think it's way faster, but not
sure about that. Also used new syntax in changed parts of build
process. Node.js has improved filesystem methods over time,
so I made use of them.

Changes:
 - Removed `trim` module.
 - Wrapped `execSync` and used `{ encoding: "utf-8" }` everywhere.
 - Simplified `fs.readFileSync` to specify encoding as second argument, instead of creating a buffer
 - Got rid of `version` file, instead specifying `version` in `package.json` and reading it.

I believe this will make build at least 0.01 seconds faster. :smile:
P.S. Updated GitHub CLI right before creating PR... it broke again
<details>
    <summary>Log</summary>
<pre>
? Which should be the base repository (used for e.g. querying issues) for this directory? tobspr/shapez.io

Creating pull request for i-hate-trim-module into master in tobspr/shapez.io

? Title Simplify build process
? Body <Received>
? What's next? Submit
pull request create failed: GraphQL error: Head sha can't be blank, Base sha can't be blank, No commits between master and i-hate-trim-module, Head ref must be a branch
</pre>
</details>
As you can see, there IS one commit (and creating PR worked earlier)


tobspr:
Didn't get to try it yet

Issue #751, opened 2020-10-03
setting keybinding . displays as ¾


shawjef3:
I set a keybinding to period but it displays as ¾.

EmeraldBlock:
It seems to be a fault of getStringForKeyCode (period is key code 190, but String.fromCharCode thinks that's ¾)

PR #753, opened 2020-10-03
"Fix" getStringForKeyCode returning incorrect strings

EmeraldBlock:
This adds the full stop/period `.` to the switch statement. This also modifies the function executed if the key code is not found so that it will return the key code surrounded by brackets if it does not correspond to a letter or a digit key. (For example, the grave accent <code>`</code> returns ``[192]``. Keys like <code>2</code> and <code>F</code> pass normally, but not the numpad keys.)

This is definitely not a great fix, but it should prevent keys not specified from displaying incorrectly. Maybe using <code>�</code> would be better?
image
(Fixes issue #751)

### DIFF
--- a/src/js/game/key_action_mapper.js
+++ b/src/js/game/key_action_mapper.js
@@ -262,7 +262,9 @@ export function getStringForKeyCode(code) {
             return "'";
     }
 
-    return 48 <= code && code <= 57 || 65 <= code && code <= 90 ? String.fromCharCode(code) : "[" + code + "]";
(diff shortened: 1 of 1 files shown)


tobspr:
Alright, well its a good first solution, I'll think about something better! :)

pepkin88:
I think using the `key` field instead of the deprecated `keyCode` would give better results. It already returns a human-readable label of a pressed key.

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key

PR #756, opened 2020-10-04
Added 75, 144 and 165hz refresh rates

KaiSforza:
These are pretty common refresh rates for most monitors, 75 being old
and 165 being somewhat new, but used reasonably often in higher-end
monitors.

I was using the 144 and when I went to the beta it was no longer there.


tobspr:
They were removed for a reason, they are actually no longer called refresh rates but tick rates

PR #757, opened 2020-10-04
Fix yaml errors in CI

KaiSforza:
Don't warn on document start (cleanup, no one is using this)

Remove a bunch of trailing spaces and fix some bad indentation in the
pt-PT file.

There are still a bunch of lines that are significantly too long, should
be multiline strings.


tobspr:
Awesome, thanks! :)

PR #758, opened 2020-10-04
Run yamllint in parallel

KaiSforza:
This doesn't require yarn to build anything or ffmpeg to be ready, so
just do this in parallel with the rest of the jobs.

Feedback would be awesome. I don't know if this is going to run properly.


tobspr:
Awesome, thank you! :)

tobspr:
I'll simply give it a try

PR #768, opened 2020-10-05
Improved Dockerfile performance + added java dependency

hrosseel:
Included java "default-jre" dependency in the Dockerfile.


tobspr:
Thanks! :)

PR #773, opened 2020-10-05
Added Arabic translation

mohaemen:


tobspr:
Thanks! I'm interested if it will work within a LTR game :P

PR #776, opened 2020-10-06
Prevent insta-crash when screenshotting larger bases

EmeraldBlock:
This prevents the game from trying to access an atlas link of nonexistent resolution. Whenever a base is larger than 128 chunks long or wide, the `chunkScale` value is set to below 0.25, the smallest atlas scale. This changes it so that the atlas scale used is always 0.25.

I really don't know how the atlas works that well, though I have tested this on my large base from the standalone version (producing an image of ~30.2 MB). The image (downscaled 4x on each axis) is below:
basebigsmall


tobspr:
Ah, awesome, thank you! Yeah I should fix it

PR #785, opened 2020-10-07
Multi Layer Selecting

CEbbinghaus:
A Small QOL Improvement allowing Multiple layers to be selected at once. by adding the Alt key they can now select both wires and regular blocks and copy/cut & paste them together. this works in combination with the Shift Multi Area copy.

Copy-Paste


src/js/game/hud/parts/mass_selector.js
```diff
@@ -243,9 +244,18 @@ export class HUDMassSelector extends BaseHUDPart {
...
-                    const contents = this.root.map.getLayerContentXY(x, y, this.root.currentLayer);
-                    if (contents && this.root.logic.canDeleteBuilding(contents)) {
-                        this.selectedUids.add(contents.uid);
+                    let entities = [];
+                    if (
+                        this.root.keyMapper.getBinding(KEYMAPPINGS.massSelect.massSelectSelectMultiLayer)
+                            .pressed
+                    )
+                        entities = this.root.map.getLayersContentsMultipleXY(x, y);
+                    else entities = [this.root.map.getLayerContentXY(x, y, this.root.currentLayer)];
```
tobspr, line 253: I always use braces (`{}`), dunno why the linter didn't complain, but I can change that myself :)

src/js/game/hud/parts/mass_selector.js
```diff
@@ -288,48 +298,83 @@ export class HUDMassSelector extends BaseHUDPart {
             parameters.context.fill();
             parameters.context.stroke();
 
-            parameters.context.fillStyle = THEME.map.selectionOverlay;
-
             const renderedUids = new Set();
 
+            let isMultiLayerPressed = this.root.keyMapper.getBinding(
```
tobspr, line 303: `const`

src/js/game/hud/parts/mass_selector.js
```diff
@@ -243,9 +244,18 @@ export class HUDMassSelector extends BaseHUDPart {
...
+                    if (
+                        this.root.keyMapper.getBinding(KEYMAPPINGS.massSelect.massSelectSelectMultiLayer)
+                            .pressed
+                    )
+                        entities = this.root.map.getLayersContentsMultipleXY(x, y);
+                    else entities = [this.root.map.getLayerContentXY(x, y, this.root.currentLayer)];
+
+                    for (let i = 0; i < entities.length; ++i) {
+                        let entity = entities[i];
+                        if (entity !== null && this.root.logic.canDeleteBuilding(entity))
```
tobspr, line 257: It's enough to do `if (entity && ....)`

tobspr (changes requested):
So, I like the idea, but there are a few things:

1) Code style (But I can fix that myself)
2) The code for drawing a selected entity should be probably moved into a separate method, since its now duplicated and contains a lot of logic
3) The selection is already slow to draw, with the `drwaSingleWiresOverviewTile` it will be *incredibly* slow because that draws up to 9 squares. So first, we would have to find a better solution on how to draw the selection, because otherwise it gets unusable

src/js/game/hud/parts/mass_selector.js
CEbbinghaus: 👍

src/js/game/hud/parts/mass_selector.js
CEbbinghaus: 👍

src/js/game/hud/parts/mass_selector.js
CEbbinghaus: will change

CEbbinghaus:
will fix formatting errors and move the code to a separate function. we can use render it all as squares for now but in my experience its much better to visually see what you are selecting. reworking the layer preview code to allow for rectangular previews might help to see what exactly you are previewing. 

i have also found that while the alt key works well to allow to user selection of the selection mode i tend do let go just as i release the mouse and only copy one layer. maybe a grace period in which the key stays pressed after having let go or alternatively using alt to toggle the multi layer might be a good idea.

CEbbinghaus:
Fixed the Problems and temporarily changed the rendering to only render squares.
image

Maybe rendering the Blueprint sprites like when pasting might be a better alternative.

CEbbinghaus:
Reused the Blueprint rendering code because it looks much better and gives a good view into the wire layer when selecting:
image

tobspr:
Can you have a look at the linter issues again? :D

CEbbinghaus:
there arent any in the latest commits?
What do you mean?

tobspr:
It's on my todo list

src/js/game/hud/parts/mass_selector.js
```diff
@@ -288,49 +309,61 @@ export class HUDMassSelector extends BaseHUDPart {
...
-                bounds.w * globalConfig.tileSize - 2 * boundsBorder,
-                bounds.h * globalConfig.tileSize - 2 * boundsBorder,
-                2
-            );
-            parameters.context.fill();
-        });
+		//EXTREMELY SLOW. There must be a better way. (Possibly use a Array)
+		for(let i = 0; i < this.selectedEntities.length; ++ i){
+			const entity = this.selectedEntities[i];
+			this.RenderSelectonPreviewTile(parameters, entity);
```
jw2476, line 343: I've looked over some of the other components and it seems fairly normal to render things in this way.

jw2476 (approved):

tobspr:
Could you please fix the code style first?
1. Proper indentation
2. Naming (camelCase for methods)
3. Eslint errors

CEbbinghaus:
Sorry, I have been gone for so long. I am picking this back up and will fix any and all outstanding Issues.
@tobspr Will update and merge all changes, then fix Naming and any ESlint errors. I will also fix the indentation and verify that it's working and functional.

src/js/game/core.js
```diff
@@ -425,7 +425,12 @@ export class GameCore {
 
         root.camera.transform(context);
 
-        assert(context.globalAlpha === 1.0, "Global alpha not 1 on frame start");
+        if (context.globalAlpha !== 1.0) {
+            console.warn("Global Alpha not set back to 1 on Frame Begin");
+            context.globalAlpha = 1;
+        }
+
+        // assert(context.globalAlpha === 1.0, "Global alpha not 1 on frame start");
```
CEbbinghaus, line 433: I'm going to go out on a limb and assume that this is against what you want. Personally, I found that it kept happening and I couldn't track down where the issue was originating. Since changing it over to the if statement I haven't had any issues and I feel like it's a better solution overall. Notifying the developer of the problem but fixing it so it doesn't have effects.

src/js/game/hud/parts/mass_selector.js
```diff
@@ -155,13 +148,12 @@ export class HUDMassSelector extends BaseHUDPart {
...
     clearBelts() {
-        for (const uid of this.selectedUids) {
-            const entity = this.root.entityMgr.findByUid(uid);
+        for (const entity of this.selectedEntities) {
             for (const component of Object.values(entity.components)) {
                 /** @type {Component} */ (component).clear();
             }
         }
-        this.selectedUids = new Set();
+        this.selectedEntities = [];
```
CEbbinghaus, line 156: This is the only place selectedUid's were used so it feels like it might be leftover from my branches changes. Moved everything to the selectedEntities array and it seems to be working flawlessly

src/js/game/hud/parts/mass_selector.js
```diff
@@ -388,7 +375,7 @@ export class HUDMassSelector extends BaseHUDPart {
      * @param {DrawParameters} parameters
      * @param {Entity} entity
      */
-    RenderSelectonPreviewTile(parameters, entity) {
+    renderSelectonPreviewTile(parameters, entity) {
```
CEbbinghaus, line 378: Fixed the Naming of the function to be inline with all other camelCase methods

CEbbinghaus (commented):
Hope the changes are up to spec. if needed i can remove the if statement and restore the assert but i would want to ensure that the global alpha problem has been resolved

Sense101:
why don't you use prettier or an automatic code linter so that this won't fail the checks?

CEbbinghaus:
> why don't you use prettier or an automatic code linter so that this won't fail the checks?

I am...

PR #789, opened 2020-10-07
Fix non-uniform distribution for freeplay shape color palette

EmeraldBlock:
This fixes the issue mentioned in #747 by using the nextIntRange function instead of the nextIntRangeInclusive function.
The nextIntRangeInclusive function is also removed, since it gives the minimum and maximum integer half as much weight. It is not used elsewhere.

tl;dr does what i said in that other pr


tobspr:
Ah, cool , that has been on my todo list!

PR #790, opened 2020-10-08
Allow for cycling buildings and variants backwards

EmeraldBlock:
This makes cycling buildings or variants go backwards if the Rotate CCW Modifier key (default SHIFT) is pressed. This has the consequence of changing the meaning of the modifier key.


src/js/game/hud/parts/base_toolbar.js
```diff
@@ -160,9 +160,16 @@ export class HUDBaseToolbar extends BaseHUDPart {
         }
 
         let newBuildingFound = false;
+        const direction = this.root.keyMapper.getBinding(KEYMAPPINGS.placement.rotateInverseModifier)
```
tobspr, line 163: Linter issue

src/js/game/hud/parts/base_toolbar.js
```diff
@@ -160,9 +160,16 @@ export class HUDBaseToolbar extends BaseHUDPart {
...
 
         let newBuildingFound = false;
+        const direction = this.root.keyMapper.getBinding(KEYMAPPINGS.placement.rotateInverseModifier)
+            .pressed
+            ? -1
+            : 1;
         let newIndex = this.lastSelectedIndex;
-        for (let i = 0; i < this.primaryBuildings.length; ++i, ++newIndex) {
+        for (let i = 0; i < this.primaryBuildings.length; ++i, newIndex += direction) {
             newIndex %= this.primaryBuildings.length;
```
tobspr, line 169: `newIndex = safeModulo(newIndex + direction)` and remove the `++newIndex` from the for loop or ?

src/js/game/hud/parts/building_placer_logic.js
```diff
@@ -467,7 +467,14 @@ export class HUDBuildingPlacerLogic extends BaseHUDPart {
                 index = 0;
                 console.warn("Invalid variant selected:", this.currentVariant.get());
             }
-            const newIndex = (index + 1) % availableVariants.length;
+            const direction = this.root.keyMapper.getBinding(KEYMAPPINGS.placement.rotateInverseModifier)
+                .pressed
+                ? -1
+                : 1;
+            let newIndex = (index + direction) % availableVariants.length;
```
tobspr, line 474: see above.

Also don't use `rotateInverseModifier` -> Instead lets add a new bindign

tobspr (changes requested):

src/js/game/hud/parts/building_placer_logic.js
Vlad1slaw: Another key? Omg...
Same feature for Cycle buildings("Tab")?
So boring to cross half of screen, just to pick trash can...
Or ability to use scroll wheel with an extra binding(ofcource different from one used here :) )

EmeraldBlock: I personally feel that adding another binding would be excessive - I've always treated `(E ->) TAB -> T -> R` as a group of bindings that let you navigate the building options, and I think a single `backwardsModifier` or something would make the most sense

tobspr: Yeah I think we should unify it into one binding

EmeraldBlock: Should rotateInverseModifier be changed to something else, and/or should the current translations of it be reverted?

tobspr: I'd:
1) Get rid of rotateInverseModifier
2) Introduce a new binding "Backwards modifier"

tobspr:
Need to have a look

tobspr:
I redid this PR because there were some issues and it would have broken all upcoming PR's regarding translations

PR #793, opened 2020-10-08
Update base-zh-TW.yaml

Colleabois:
Update some translation that are already done in old patches.
Major modification: The translation of the word "Standalone". I believe 單機 is better translation than 獨立 for describing a game that can run individually on one PC without internet connection.


tobspr:
Awesome, thank you! :)

PR #798, opened 2020-10-08
Updating yaml check again

KaiSforza:


DerFeldspatz:
I have the same changes in #794 
Only one needs to be merged

KaiSforza:
Fixed in #794

PR #801, opened 2020-10-08
Modified translations, formatting and order

ZrubekLukas:
Reviewed first 273 lines of file. Formatting and order was changed acc. to EN version for easier comparison. Added and translated comments from EN. Started list of rules (file start) for consistent future translations.

### DIFF
--- a/translations/base-cz.yaml
+++ b/translations/base-cz.yaml
@@ -1,62 +1,99 @@
+#
+# PŘEKLADY HRY
+#
+#  Přispívání:
+#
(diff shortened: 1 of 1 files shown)


ZrubekLukas:
CI / yaml-lint check is not passed because of too long lines and trailing spaces. The lines are roughly same length as in English version. Should I wrap them or leave it? Tomorrow I will check the trailing spaces.

tobspr:
Awesome, thank you! :)

PR #807, opened 2020-10-09
Update base-tr.yaml

dobidon:
- Translating new keys
- Replacing reverted translations at lines:
  688-690
  712-716
  736-742


tobspr:
Thank you!

PR #810, opened 2020-10-09
Fix reader maximum displayed speed

EmeraldBlock:
This sets the maximum reader value to the correct number. It will reflect changes in the belt's base speed but will not reflect changes in the maximum upgrade multiplier.


tobspr:
Thanks!

tobspr:
Should probably be more dynamic, but for now it works

PR #817, opened 2020-10-09
Improved base-en.yaml translations for tips

Hyperion-21:
Corrected spelling, grammar, and more rarely any wrong information for the english translation

### DIFF
--- a/translations/base-en.yaml
+++ b/translations/base-en.yaml
@@ -1209,7 +1209,6 @@ tips:
     - You don't need to divide up items evenly for full efficiency.
     - Holding <b>SHIFT</b> will activate the belt planner, letting you place long lines of belts easily.
     - Cutters always cut vertically, regardless of their orientation.
-    - To get white, mix all three colors.
     - The storage buffer prioritises the left output.
     - Invest time to build repeatable designs - it's worth it!
     - Holding <b>SHIFT</b> lets you place multiple buildings in one click.
@@ -1231,8 +1230,9 @@ tips:
     - Make a separate blueprint factory. They're important for modules.
     - Have a closer look at the color mixer, and your questions will be answered.
     - Use <b>CTRL</b> + Click to select an area.
-    - Building too close to the hub can get in the way for later projects.
+    - Building too close to the hub can get in the way of later projects.
     - The pin icon next to each shape in the upgrade list pins it to the screen.
+    - Mix all three primary colors to make white!
     - You have an infinite map. Don't cramp your factory, expand!
     - Also try Factorio! It's my favorite game.
     - The quad cutter cuts clockwise, starting from the top right.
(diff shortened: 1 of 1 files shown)


translations/base-en.yaml
```diff
@@ -1209,38 +1209,37 @@ tips:
     - You don't need to divide up items evenly for full efficiency.
     - Holding <b>SHIFT</b> will activate the belt planner, letting you place long lines of belts easily.
     - Cutters always cut vertically, regardless of their orientation.
-    - To get white mix all three colors.
+    - To get white, mix all three colors.
     - The storage buffer prioritises the left output.
     - Invest time to build repeatable designs - it's worth it!
-    - Holding <b>CTRL</b> allows to place multiple buildings.
+    - Holding <b>SHIFT</b> lets you place multiple buildings in one click.
```
EmeraldBlock, line 1215: I think that refers to blueprints, but your version changes that to multiplace

translations/base-en.yaml
```diff
@@ -1209,38 +1209,37 @@ tips:
...
+    - You may need to reuse factories later in the game. Build your factories to be reusable.
     - Sometimes, you can find a needed shape in the map without creating it with stackers.
     - Full windmills/pinwheels can never spawn naturally.
     - Color your shapes before cutting for maximum efficiency.
     - With modules, space is merely a perception; a concern for mortal men.
     - Make a separate blueprint factory. They're important for modules.
     - Have a closer look at the color mixer, and your questions will be answered.
     - Use <b>CTRL</b> + Click to select an area.
-    - Building too close to the hub can get in the way of later projects.
+    - Building too close to the hub can get in the way for later projects.
```
EmeraldBlock, line 1234: I think most would use "get in the way of"; "for" makes it sort of sound like building too close and later projects work together

translations/base-en.yaml
```diff
@@ -1209,38 +1209,37 @@ tips:
...
     - The pin icon next to each shape in the upgrade list pins it to the screen.
-    - Mix all three primary colors to make white!
```
EmeraldBlock, line 1236: It might be better to take out the earlier white-mixing tip, since this one flows a little better and provides more clarification

translations/base-en.yaml
Hyperion-21: holding crtl during blueprint placement does nothing from experience

EmeraldBlock: Yeah, the old thing doesn't make much sense, though changing "in one click" to "at a time" may be better

tobspr:
Thanks! I'll do a minor adjustment to the level 26 hint

translations/base-en.yaml
VoxLight:
```suggestion
    - Holding <b>SHIFT</b> lets you drag to place buildings.
```

PR #818, opened 2020-10-09
Refactor: Entity Arrays => Sets, faster loads & copy/paste

davisgamedev:
This PR moves all instances of entity iteration by replacing entity arrays with es6 sets. This provides significant performance boosts during operations such as copy and paste, and loading/saving.


tobspr:
This is on my todo list, since its so big it will take a bit of time to review though! Thanks for the effort so far!

src/js/game/blueprint.js
```diff
@@ -57,6 +57,37 @@ export class Blueprint {
         return new Blueprint(newEntities);
     }
 
+    /**
+     * Creates a new blueprint from the given entity uids
+     * @param {Array<Entity>} entities
+     */
+    static fromEntities(entities) {
+        const newEntities = [];
```
elidoran, line 65: Let it know at once you want it to hold the same number of elements.

```suggestion
        const newEntities = [];
        newEntities.length = entities.length;
```

Or, use the Array constructor to do the same thing.
```suggestion
        const newEntities = new Array(entities.length);
```

src/js/game/blueprint.js
```diff
@@ -57,6 +57,37 @@ export class Blueprint {
...
+
+        let averagePosition = new Vector();
+
+        // First, create a copy
+        for (let i = entities.length - 1; i >= 0; --i) {
+            const entity = entities[i];
+
+            const clone = entity.clone();
+            newEntities.push(clone);
```
elidoran, line 74: Then, assign the value using `i` instead of using `push(clone)` because push spends time handling if multiple arguments were provided to add to the array.

```suggestion
            newEntities[i] = clone;
```

elidoran (commented):
Just a couple of suggestions I saw when looking over all the changes.

src/js/game/blueprint.js
davisgamedev: Thanks this is a good suggestion!

davisgamedev:
Hi! This branch has been updated to address comments and keep up to date with upstream changes

PR #819, opened 2020-10-09
yaml-lint fixes

davisgamedev:
Added line breaks to most yaml translations. Fixed typos in en. Fixed error in ar.

No translation verbiage or formats have been changed, but would love a double check to see if that is the case.

Oh my god the yaml linter args could've fixed this too couldn't've. Oh well.


tobspr:
I'll fix the linter instead :P

PR #823, opened 2020-10-10
Translated ~90% of the strings to spanish

Galo223344:
There are some strings left and the tips and hints are not translated.


tobspr:
Thank you!

PR #828, opened 2020-10-10
Make reward notification scrollable

LeopoldTal:
Some level rewards have a long description. This can push the "Next level" button below the screen. The user is then stuck and can't proceed without relaunching the game.

Level 20 in German overflows the screen, hiding the button

This PR allows the user to scroll and reveal the button.


tobspr:
Ah, thanks! :)

PR #830, opened 2020-10-10
Format all upgrade tiers as Roman numerals

LeopoldTal:
Currently, upgrade tiers up to 20 are formatted as Roman numerals, but fall back to Arabic numerals from tier 21.

Tier 20 is shown as Tier XX, but not tier 21

This PR adds Roman-numeral formatting up to the limit of the standard system (3999), which covers all 1000 tiers.

Tier 388 shown as Tier CCCLXXXVIII

### DIFF
--- a/src/js/core/utils.js
+++ b/src/js/core/utils.js
@@ -714,7 +714,7 @@ export function startFileChoose(acceptedType = ".bin") {
     });
 }
 
-const MAX_ROMAN_NUMBER = 3999;
+const MAX_ROMAN_NUMBER = 49;
 const romanLiteralsCache = ["0"];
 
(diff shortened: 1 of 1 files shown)


Dimava:
IMAO you wouldn't want to have romanic numbers for numbers over 50
they are harder to understand that it's needed

tobspr:
What @Dimava said

LeopoldTal:
Ok, limited it to 49 = `XLIX`. Left the general method, easier to change later.

tobspr:
Not too sure about this, gonna have to have some thoughs

tobspr:
Thanks!

Issue #831, opened 2020-10-10
[Bug] Clear cursor on right click doesn't work correctly when over a button.


olllie-SC:
When right clicking to clear your cursor while hovering over a button it deletes the placed button and clears your cursor at the same time.

![shapez button bug](https://user-images.githubusercontent.com/72669480/95663504-b5f2ca00-0b0d-11eb-8a34-67a4ed663504.gif)

PR #832, opened 2020-10-10
Fix button right click bug

artemisSystem:
Previously, right clicking a button with an item selected would delete the button (instead of just deselect the item in hand). Looks like it was because the button right click had its own handler. Removing it seems to have fixed the issue.

### DIFF
--- a/src/js/game/hud/parts/layer_preview.js
+++ b/src/js/game/hud/parts/layer_preview.js
@@ -76,6 +76,12 @@ export class HUDLayerPreview extends BaseHUDPart {
                 const tileY = dy + startTileY;
 
                 const content = this.root.map.getLayerContentXY(tileX, tileY, "wires");
+                const contentRegular = this.root.map.getLayerContentXY(tileX, tileY, "regular");
+                if (contentRegular && content) {
+                    // We don't want to count something on the wires layer here if it is on the normal layer,
+                    // because then it will get rendered twice
+                    continue;
+                }
                 if (content) {
                     MapChunkView.drawSingleWiresOverviewTile({
                         context: this.context,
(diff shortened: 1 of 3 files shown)


3ddyy:
Fixes #831

3ddyy:
(Looks like that didn't link the issue successfully, looks like tobias can link it in the sidebar, i cant)

tobspr:
This will make it impossible to delete buttons from the wires layer though, not?

3ddyy:
Ah, yes, you're correct. It can probably be fixed by adding back the lines i removed and adding an extra check, but i thought it would probably be better to do this along with where other deletion on the wires layer is handled (that way it would also be easy to let the storage, belt reader, etc be deletable from the wires layer too). But! I couldn't find the code for that anywhere. Could you point me in the direction of it and i'll see what i can do? I could instead just add a simple check in the `lever_toggle.js`, if you think that would be better.

tobspr:
Sorry for the slow response, I've been super busy

The deletion itself is happening in `building_placer_logic.js` in `deleteBelowCursor`, is that what you mean?

artemisSystem:
Thanks! I found what i was looking for. I'll try to make it so that if there is nothing on a tile in the wires layer, then if there is a button, reader, or display on the regular layer, then `this.root.map.getTileContent(tile, "wires")` will return that component (making it get deleted, pipette'd, etc). I don't think it's a good idea for filter or storage, since wires can be placed on top of those. I looked through all the uses of `getLayerContentFromWorldCoords`, and it doesn't seem like this will break anything with some few changes.

artemisSystem:
There! It seems to work like a charm.

tobspr:
I don't like the change to the map chunk, it is a pretty big change - instead I think the logic which controls the deletion etc should get smarter, no?

tobspr:
Closing due to no response, feel free to reopen

PR #837, opened 2020-10-11
Update base-ind.yaml

presdm:
Update on tips, matching it with the new improved base version. Also corrected some things.

- PresS#9257


tobspr:
Thanks! :)

PR #838, opened 2020-10-11
Update base-cz.yaml

RevosCZ:
Corrected one of the tips. ZrubekLukas also added some further corrections and translations.


tobspr:
Merged in #918

PR #872, opened 2020-10-13
Fix failing build

Mariusmivw:
The failing kind of annoyed me so I used Gravis-CI (https://github.com/DanySK/Gravis-CI) to fix the travis build error
I also removed the trailing spaces in some translation files


tobspr:
There are conflicts, can you have a look?

Mariusmivw:
Should be fixed now

Issue #842, opened 2020-10-11
Double Painter performance below advertised


vanyamil:
Here is the setup. 12 input circles, 12 input green dye. Upgrades are Belts 5 (12 / s), Paint 4 (double painter advertised as 1/s for each input, aka 2/s of output).

![image](https://user-images.githubusercontent.com/16938366/95678127-918fff80-0b98-11eb-8dda-4b8d904b7175.png)

Performance at about 95% (5.7 average of 6).

ktsimpso:
I have the same issue. Tried adjusting to all tick rates of 60, 120, and 240 but no change in behavior. Regular painter seems to work as expected.

![painter_bug](https://user-images.githubusercontent.com/5854382/95800171-9ee9de80-0cab-11eb-9400-c8b312d490b0.png)

tobspr:
known bug

PR #882, opened 2020-10-14
Update item_processor.js

generalcrazy:
Fixes https://github.com/tobspr/shapez.io/issues/842
Fix for delays in processor charge time, pervents processing delays when having more outputted items then ejectors (double painter) and delays in update ticks

Belt Speed x8(Tier VIII), all others x8.10(Tier IX)
Before
image

After
image

### DIFF
--- a/src/js/game/systems/item_processor.js
+++ b/src/js/game/systems/item_processor.js
@@ -76,16 +76,23 @@ export class ItemProcessorSystem extends GameSystemWithFilter {
 
             const currentCharge = processorComp.ongoingCharges[0];
 
-            if (currentCharge) {
-                // Process next charge
-                if (currentCharge.remainingTime > 0.0) {
-                    currentCharge.remainingTime -= this.root.dynamicTickrate.deltaSeconds;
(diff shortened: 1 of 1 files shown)


guru-coding (changes requested):
Basically this patch suggests to apply **additional bonus time** when there are more items than we can eject in one iteration for some reason (double painter issue). That could be done almost in one line.

Add **ELSE** statement after the line
https://github.com/tobspr/shapez.io/blob/f8371a96cf02e988e9283d06adbc9a21aceac91e/src/js/game/systems/item_processor.js#L87
```js
else {
     // Add more bonus time if charge finished but we still have items to eject
     processorComp.bonusTime += this.root.dynamicTickrate.deltaSeconds;
}
```

We **don't need** to apply bonus time directly here, because bonus time **will be** applied for a new charge there
https://github.com/tobspr/shapez.io/blob/f8371a96cf02e988e9283d06adbc9a21aceac91e/src/js/game/systems/item_processor.js#L297

tobspr:
What @guru-coding said

PR #887, opened 2020-10-15
Update base-nl.yaml

chris-kruining:
global.suffix.thousands :: should be uppercase as is stands for Kilo, which is shortened to K
global.infinite :: `oneindig` is the translation for infinite, not sure if keeping it as `inf` was on purpose.
global.time.* :: consistency
mainMenu.changelog :: -
mainMenu.continue :: technically a more correct translation
dialogs.buttons.restart :: consistency
dialogs.gameLoadFailure.title :: seems better to me

I will try to translate some more in the future, but my patience has run out at this moment :P


translations/base-nl.yaml
```diff
@@ -97,23 +97,23 @@ global:
...
     suffix:
-        thousands: k
+        thousands: K
         millions: M
         billions: B
         trillions: T
 
     # Shown for infinitely big numbers
-    infinite: inf
+    infinite: oneindig
```
DerFeldspatz, line 106: This might be too long. Use a shortening or infinity sign

translations/base-nl.yaml
```diff
@@ -97,23 +97,23 @@ global:
 
     # The suffix for large numbers, e.g. 1.3k, 400.2M, etc.
     suffix:
-        thousands: k
+        thousands: K
```
DerFeldspatz, line 100: You could also switch to SI (k,M,G,T) if there are no shortenings in dutch

chris-kruining: there is a shortening, and as I explained, should be an uppercase K, not lowercase

DerFeldspatz: Okay

translations/base-nl.yaml
chris-kruining: as in the utf8 char for infinity? `∞` (U+221E)

DerFeldspatz: Yeah, you can use it if you want

tobspr:
Could you have a look at the conflicts? :)

chris-kruining:
@tobspr done

translations/base-nl.yaml
```diff
@@ -117,8 +117,8 @@ dialogs:
         title: Savegame geïmporteerd
         text: Je savegame is succesvol geïmporteerd.
     gameLoadFailure:
-        title: Het spel is kapot
-        text: "Het laden van je savegame is mislukt:"
+        title: Corrupte savegame
+        text: Het laden van je savegame is mislukt:
```
ZrubekLukas, line 121: The quotes should stay at line 121 otherwise you get the lint-check error.

chris-kruining: done

tobspr:
Ty!

PR #888, opened 2020-10-15
Make save compressor faster

### LINKED ISSUE #836: Low performance when saving with lots of shapes delivered
So uhh... I built an auto everything machine. Thing is it uhh... delivered a little bit of everything... 250k different shapes.
Now for the most part the game seems to handle that ok, but saving takes 300s or so.

The issue seems to be the compressObject method.

In specific line 69-81 of `/js/savegame/savegame_compressor.js`

```
    } else if (typeof obj === "object" && obj !== null) {
        let result = {};
        for (const key in obj) {
            let index = keys.indexOf(key);
            if (index < 0) {
                keys.push(key);
                index = keys.length - 1;
            }
            const value = obj[key];
            result[compressInt(index)] = compressObjectInternal(value, keys, values);
        }
        return result;
    }
```

In specific keys.indexOf(key) is **super** slow if you have 250k keys in that array. Sample of the keys array... every shape is dumped into it...
        "CrCgSuSu:SuWySuSy:SbSbWgRr:WuWuSgRg",
        "CgCgSuSu:WbWbSrSy:SbSbRgRg:WuWuCcCc",
        "CgCgSuSu:WbWbCrCg:SbSrRgRr:WuWuSgRg",
        "CgCgSuSu:WbWbSuSy:SgSbWgRr:WuWuCcCc",
        "CgCgSuSu:WbSySuSy:WbSrWgRr:WuWuCcCc",
        "CrCgSuSu:WbWbCgCg:SbSbWgRr:WuWuCcCg",
        "CgCgSuSu:SuWySuSy:SbSbRgRg:WuWuCcCg",
        "CgCgSuWu:WbWbCgCg:SbSbRgRg:WuWuSgRg",
        "CrCgSuSu:SbWySuSy:SbSbRgRg:WuWuCcCc",
        "CgCgSuSu:WbWbCgCg:SbSbRgRg:WuWuCcCc",
        "CgCgSuSu:WbWbCrCy:SbSbWgRr:WuWuCgCg",
        "CrCgSuSu:WbWySuSy:SbSbRgRg:WuWuCcCc",
        "CgCgSuSu:WbWbCgCg:SbSbRgRr:WuWuSgRg",
        "CgCgSuSu:WbWbCrSy:SbSbRgRr:WuWuCcCc",

I was thinking something like this as a solution, though it's not quite right and I don't really like it. An off the shelf compression library would probably be a better solution.

```
function compressObjectInternal2(obj, keyCount = 0, keys = [], values = []) {
    if (Array.isArray(obj)) {
        let result = [];
        for (let i = 0; i < obj.length; ++i) {
            result.push(compressObjectInternal2(obj[i], keyCount, keys, values));
        }
        return result;
    } else if (typeof obj === "object" && obj !== null) {
        let result = {};
        for (const key in obj) {
            let index = keys[key];
            if (!index) {
                keys[key] = keyCount;
                index = keyCount;
                keyCount += 1;
            }
            const value = obj[key];
            result[compressInt(index)] = compressObjectInternal2(value, keyCount, keys, values);
        }
        return result;
    } else if (typeof obj === "string") {
        let index = values.indexOf(obj);
        if (index < 0) {
            values.push(obj);
            index = values.length - 1;
        }
        return compressInt(index);
    }
    return obj;
}

export function compressObject2(obj) {
    const keys = {};
    const values = [];
    const data = compressObjectInternal2(obj, 0, keys, values);
    return {
        keys: Object.keys(keys),
        values,
        data,
    };
}
```

For anyone encountering this problem that needs a quick patch I just hacked this into the save section
```
export function compressObject(obj) {
    const keys = [];
    const values = [];
    if (obj && obj.dump && obj.dump.hubGoals && obj.dump.hubGoals.storedShapes) {
        const storedShapes = obj.dump.hubGoals.storedShapes;
        for (let key in storedShapes) {
            if (storedShapes[key] < 100) {
                delete storedShapes[key];
            }
        }
    }
    const data = compressObjectInternal(obj, keys, values);
    return {
        keys,
        values,
        data,
    };
}
```

Then loaded and saved my game again.

LeopoldTal:
Fixes #836 by using a Map instead of an array.

The Map maps the values to their indices rather than the reverse, making finding the index constant-time rather than linear, and so building the full map linear rather than quadratic. Flipping it to build the final array is also linear.

On my machine, improves save time on @macdja38's test save from >5 minutes to about 30 seconds.

### DIFF
--- a/src/js/savegame/savegame_compressor.js
+++ b/src/js/savegame/savegame_compressor.js
@@ -59,7 +59,7 @@ if (G_IS_DEV) {
     }
 }
 
-function compressObjectInternal(obj, keys = [], values = []) {
+function compressObjectInternal(obj, keys, values) {
     if (Array.isArray(obj)) {
         let result = [];
         for (let i = 0; i < obj.length; ++i) {
@@ -69,33 +69,41 @@ function compressObjectInternal(obj, keys = [], values = []) {
     } else if (typeof obj === "object" && obj !== null) {
         let result = {};
         for (const key in obj) {
-            let index = keys.indexOf(key);
-            if (index < 0) {
-                keys.push(key);
-                index = keys.length - 1;
+            let index = keys.get(key);
+            if (index === undefined) {
+                index = keys.size;
+                keys.set(key, index);
             }
             const value = obj[key];
             result[compressInt(index)] = compressObjectInternal(value, keys, values);
         }
         return result;
     } else if (typeof obj === "string") {
-        let index = values.indexOf(obj);
-        if (index < 0) {
-            values.push(obj);
-            index = values.length - 1;
+        let index = values.get(obj);
+        if (index === undefined) {
+            index = values.size;
+            values.set(obj, index);
         }
         return compressInt(index);
     }
     return obj;
 }
 
+function indexMapToArray(hashMap) {
+    const result = [];
+    hashMap.forEach((index, key) => {
+        result[index] = key;
+    });
+    return result;
+}
+
 export function compressObject(obj) {
-    const keys = [];
-    const values = [];
+    const keys = new Map();
+    const values = new Map();
     const data = compressObjectInternal(obj, keys, values);
     return {
-        keys,
-        values,
+        keys: indexMapToArray(keys),
+        values: indexMapToArray(values),
         data,
     };
 }


dengr1065:
<insert message about polyfill mess, even though it works on all browsers>. caniuse-lite is outdated.,

pavel-voronin (approved):

tobspr:
NIce, thank you!

tobspr:
PS:in the index map to array, it might make sense to initialize the array first. I'll change it

PR #897, opened 2020-10-16
Use electron builtin to open links

LeopoldTal:
When opening a link on standalone, use [shell.openExternal](https://github.com/electron/electron/blob/master/docs/api/shell.md#shellopenexternalurl-options).

This works cross-platform, and looks less suspicious to antivirus than explicitly shelling out. Users have reported Avast antivirus quarantining the game when the viewer link is opened.


tobspr:
Ah nice, thanks!

PR #898, opened 2020-10-17
Update base-it.yaml

BananoGamer:
Uploaded the new text relative to the wire layer and other small corrections, it's 100% done


tobspr:
Merged in #918

PR #913, opened 2020-10-19
Update for base-ro.yaml

MemelDanzig:
I found some words like Roadmap and Standalone to be better suited when they're in English, I think some neologisms are better kept in their language.


tobspr:
Merged in #918

PR #915, opened 2020-10-20
Fix long lines and trailing spaces

AdJaGu:
Remove trailing spaces and break up long lines.


ZrubekLukas:
Long lines and trailing spaces in base-cz are already fixed in PR #838

PR #936, opened 2020-10-26
Fix windmill shape generation for more diverse windmill resources on the map

noctilucentgames:
Going from a lot of these: 
image
To more of these:
image
image

This will slightly increase the number of windmill sub shapes on the map.

Signed-off-by: noctilucentgames <noctilucentgames@gmail.com>

### DIFF
--- a/src/js/game/map_chunk.js
+++ b/src/js/game/map_chunk.js
@@ -195,38 +195,59 @@ export class MapChunk {
         if (distanceToOriginInChunks < 7) {
             // Initial chunks can not spawn the good stuff
             weights[enumSubShape.star] = 0;
-            weights[enumSubShape.windmill] = 0;
+            weights[enumSubShape.windmill] = 0; 
+        }
+
+        if (distanceToOriginInChunks < 10) {
+            // Initial chunk patches always have the same shape
+            const subShape = this.internalGenerateRandomSubShape(rng, weights);
+            subShapes = [subShape, subShape, subShape, subShape];
+        } else if (distanceToOriginInChunks < 15) {
+            // Later patches can also have mixed ones
+            const subShapeA = this.internalGenerateRandomSubShape(rng, weights);
+            const subShapeB = this.internalGenerateRandomSubShape(rng, weights);
+            subShapes = [subShapeA, subShapeA, subShapeB, subShapeB];
+        } else {
+            // Finally there is a mix of everything
+            subShapes = [
+                this.internalGenerateRandomSubShape(rng, weights),
+                this.internalGenerateRandomSubShape(rng, weights),
+                this.internalGenerateRandomSubShape(rng, weights),
+                this.internalGenerateRandomSubShape(rng, weights)
+            ];
         }
 
         // Makes sure windmills never spawn as whole
-        let windmillCount;
-        do {
-            windmillCount = 0;
-            if (distanceToOriginInChunks < 10) {
-                // Initial chunk patches always have the same shape
-                weights[enumSubShape.windmill] = 0; // no whole windmills
+        let windmillCount = 0;
+        for (let i = 0; i < subShapes.length; ++i) {
+            if (subShapes[i] === enumSubShape.windmill) {
+                ++windmillCount;
+            }
+        }
+        if (windmillCount > 1) {
+            weights[enumSubShape.windmill] = 0;
+            if (distanceToOriginInChunks < 15) {
+                // Turn left or right subshapes into a random type that is not windmill
                 const subShape = this.internalGenerateRandomSubShape(rng, weights);
-                subShapes = [subShape, subShape, subShape, subShape];
-            } else if (distanceToOriginInChunks < 15) {
-                // Later patches can also have mixed ones
-                const subShapeA = this.internalGenerateRandomSubShape(rng, weights);
-                const subShapeB = this.internalGenerateRandomSubShape(rng, weights);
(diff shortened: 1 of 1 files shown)


EmeraldBlock:
Small suggestion: Maybe loop over the if...else if...else instead of recursively calling the shape generation function

src/js/game/map_chunk.js
```diff
@@ -198,36 +198,35 @@ export class MapChunk {
...
-        let windmillCount = 0;
-        for (let i = 0; i < subShapes.length; ++i) {
-            if (subShapes[i] === enumSubShape.windmill) {
-                ++windmillCount;
+        let windmillCount;
+        do {
+            windmillCount = 0;
+            if (distanceToOriginInChunks < 10) {
+                // Initial chunk patches always have the same shape
+                weights[enumSubShape.windmill] = 0; // no whole windmills
```
EmeraldBlock, line 207: Why set this value to `0` again? The `if (distanceToOriginInChunks < 7)` above already changes it

src/js/game/map_chunk.js
```diff
@@ -198,36 +198,35 @@ export class MapChunk {
...
-        }
-
         // Makes sure windmills never spawn as whole
-        let windmillCount = 0;
-        for (let i = 0; i < subShapes.length; ++i) {
-            if (subShapes[i] === enumSubShape.windmill) {
-                ++windmillCount;
+        let windmillCount;
+        do {
+            windmillCount = 0;
```
EmeraldBlock, line 204: Maybe move this line to where the final value is calculated since that's where it is used, though this is more an aesthetic choice

src/js/game/map_chunk.js
noctilucentgames: The distance is not increased. windmills and stars are still disabled below 7 distance like the original.

However, I added a line to disable windmills as well below 10 distance since only whole shapes can spawn and we never want whole windmill shapes. So taking out the chance of windmills avoids additional loops (since it would proc windmillCount > 2). I can of course just take out this line if you prefer.

EmeraldBlock: Oh, right, I forgot about that. Maybe just move it out since you don't have to repeatedly set it

(Not directly related to this issue, but what if 2 quads were just randomly chosen and replaced with something else? This changes the distributions, though)

noctilucentgames: The distribution of shapes will be impacted yes. I did some calculations on my current solution:
distance < 10: Windmills are removed (-100%)
distance < 15: Windmill amount is greatly increased (+83-81%)
distance > 15: Windmill amount is slightly increased (+5-8%)

The reason for this drastic change between 10 and 15 is because in the previous algorithm, when a half shape would spawn with one half windmill shape it has a 50% chance of being overridden with the rectangle, depending on if the windmill generated left or right. I assume again unintentional behaviour.

I'm not sure if you are 100% set on keeping the current balance or not. I assume the changes at <10 and <15 basically cancel each other out. From personal experience, I wouldn't mind the 5% increase at further distances at all but I understand changing balance can have severe implications and are probably not decided by me. ><

I'll make a fix that keeps the old numbers for now, but I assume a change in the actual modifiers is required instead of this.

EmeraldBlock: I honestly don't care too much about the probabilities, as long as they are somewhat close to the original amounts (and twice as many is still relatively close)

src/js/game/map_chunk.js
```diff
@@ -195,7 +195,7 @@ export class MapChunk {
         if (distanceToOriginInChunks < 7) {
             // Initial chunks can not spawn the good stuff
             weights[enumSubShape.star] = 0;
-            weights[enumSubShape.windmill] = 0;
+            weights[enumSubShape.windmill] = 0;
```
EmeraldBlock, line 198: Trailing space, was that intentional?

src/js/game/map_chunk.js
```diff
@@ -213,7 +213,7 @@ export class MapChunk {
                 this.internalGenerateRandomSubShape(rng, weights),
                 this.internalGenerateRandomSubShape(rng, weights),
                 this.internalGenerateRandomSubShape(rng, weights),
-                this.internalGenerateRandomSubShape(rng, weights),
+                this.internalGenerateRandomSubShape(rng, weights)
```
EmeraldBlock, line 216: It may be better to keep the trailing comma, to keep consistent format (also I think lint requires it)

src/js/game/map_chunk.js
```diff
@@ -225,8 +225,28 @@ export class MapChunk {
...
+                if (rng.next() < 0.5) {
+                    subShapes[0] = subShape;
+                    subShapes[1] = subShape;
+                } else {
+                    subShapes[2] = subShape;
+                    subShapes[3] = subShape;
+                }
+            } else {
+                // Turn subshape at 2 random positions into a random type that is not windmill
+                let quad1 = rng.nextIntRange(0,3);
```
EmeraldBlock, line 241: `nextIntRange()` does not include the upper bound as one of the potential values

EmeraldBlock:
Nice, and this should only affect shapes already previously affected (which may ruin some rectangle patches but whatever).

Also, you might want to resolve reviews that are no longer relevant

tobspr:
I'm not accepting this since it breaks existing savegames - resources are generated from the seed and not stored, so they will look different with this patch. It breaks the whole map generation because calling rng.nextInt for example changes the seed

PR #940, opened 2020-10-26
Many marker improvements

EmeraldBlock:
New features:

- **Waypoints can hold both shapes and text, and can hold multiple shapes.** This is done by splitting each label into shape keys and text.

- **Regular waypoints can also have compasses.** A checkbox form element is added to the marker creation dialog for this.

Other fixes/changes:

- **Two additional translations** are added for the waypoint edit notification and the compass checkbox form element.

- Text and shape width for marker boxes now correctly accounts for scale and changing scale.

- Changing the camera position is now exact, by adding a constant value onto the exponential movement.

- Waypoint sorting no longer right-pads labels with 0's.

- Some waypoint functions have changed parameters and/or return values, and some new functions are added.

- The checkbox form element has minor changes.

- The waypoint list has minor internal changes.

image


tobspr:
This is cool but give me some time to review it: ) Also would be nice not to have the translations inside since this will make the PR conflict quite often

EmeraldBlock:
welp, better fix conflicts

tobspr:
Not planning to put this into the base game, but with the 1.5.0 update it could be a mod. Closing to clean up PR's a bit.

PR #943, opened 2020-10-27
Fix processor bonus time

Dimava:


Hyperion-21:
what's this?

EmeraldBlock:
Wait a minute, doesn't `Math.max` guarantee that `timeToProcess` will be nonpositive?

tobspr:
Yeah it would be negative otherwise

PR #952, opened 2020-10-30
Visible display

TcePrepK:
I tried to make display more visible (Updated Commit Because Of Check Fails)

image


tobspr:
Closing for now since its on my todo list, there would be a lot of Changes required to the PR and it should look a bit differently :p Thanks for the effort though!

PR #968, opened 2020-11-07
Update base-zh-CN.yaml

cervonwong:
Previously it represents hours and seconds, instead of hours and minutes. (If you compare this with minutesAndSecondsShort you can tell)


tobspr:
Ty!

PR #979, opened 2020-11-09
Require LMB to place blueprints

### LINKED ISSUE #693: Binding rotate to middle click works everywhere but blueprints
I've bound rotate to middle-mouse-button, this works well for most places, but not blueprints. When I attempt to rotate blueprints, they are indeed rotated, but also pasted at the same time.

Eiim:
Fixes #693. It may be preferred to check for which mouse buttons are bound to a relevant functionality, but I think it makes more sense to just have it bound to LMB exclusively.


tobspr:
Nice, thanks!

PR #988, opened 2020-11-13
Update base-es.yaml

neftalito:
Changed "cursos" to "cursor" since is what its meant to be

### DIFF
--- a/translations/base-es.yaml
+++ b/translations/base-es.yaml
@@ -880,8 +880,8 @@ settings:
             title: Deshabilitar grilla
             description: Deshabilitar la grilla puede ayudar con el rendimiento. ¡También hace
                 que el juego se vea más limpio!
-        clearCursorOnDeleteWhilePlacing:
-            title: Limpiar el cursos al apretar click derecho
+        clearOnDeleteWhilePlacing:
+            title: Limpiar el cursor al apretar click derecho
             description: Activado por defecto, Limpia el cursor al hacer click derecho
                 mientras tengas un un edificio seleccionado. Si se deshabilita,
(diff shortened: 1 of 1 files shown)


translations/base-es.yaml
```diff
@@ -880,8 +880,8 @@ settings:
             title: Deshabilitar grilla
             description: Deshabilitar la grilla puede ayudar con el rendimiento. ¡También hace
                 que el juego se vea más limpio!
-        clearCursorOnDeleteWhilePlacing:
-            title: Limpiar el cursos al apretar click derecho
+        clearOnDeleteWhilePlacing:
```
paugay, line 883: I think you can't change this key ... correct? If you do, the translation string will not be picked up.

neftalito: yeah, the key is not the problem, the problem is "title: Limpiar el *cursor* al apretar click derecho"

EmeraldBlock: why was the key changed, then?

paugay: @neftalito I understand what you say. 

Let me explain:

You want to correct line L884 because `cursos` is wrong. You want to correct it to `cursor`, and your correction looks good. However, you can't change the key of this translation string (L883). 

If you do, the string will not get translated.

neftalito: Oh, yeah, you are right. Im sorry, didnt saw i changed the key value. My bad

paugay, line 883: This translation key still need to change 😉 .

tobspr:
Closing due to no fix provided

PR #991, opened 2020-11-15
Hide wires grid when using the "Disable Grid" setting

EmeraldBlock:
When using the "Disable Grid" setting, the wires layer grid is still visible. This is changed in this PR by replacing the repeated tile image pattern with a single color. The color is the average of the colors in the tile image and has the same transparency. The grainy look of the layer is lost.

<details><summary>"Disable Grid" OFF on left, ON on right</summary>

wireslayer
</details>
<details><summary>With buildings (confusing <b>SPOILER</b> for level 26 solution)</summary>

wireslayerbetter
</details>


tobspr:
Nice, thx!

PR #995, opened 2020-11-16
Add support for a buffered second item to be ejected

### LINKED ISSUE #842: Double Painter performance below advertised

edorfaus:
This allows the double painter to start the next paint job immediately,
thus eliminating the reason it ran slower than specified.

Fixes #842

---

This came about because i noticed that the double painter seems to be slower than specified because it waits for one of the shapes to be ejected before it starts taking in and painting the next ones.

Looking at GitHub, I found that others had tried to fix the slowdown by messing with the timing of the next paint job, but that didn't seem right to me - so I went for fixing the underlying problem instead, the waiting it does.

I envision this conceptually as the double painter having enough space in its output stage to fit both of the painted items, which is similar to how the other buildings work - they take in the next item(s) while the previous one(s) are ejected.

A small tier-1 test savefile that consistently ran too slow before (~1.8 items/sec with visible belt gaps), runs consistently at full speed (2 items/sec, no gaps) with this change in place.

Most of the implementation is in the common item ejector component, so I considered making it more generic and supporting more than one extra output item, but I figured that would be overkill since it isn't required for the double painter, and would probably make it take more CPU time.

### DIFF
--- a/src/js/game/buildings/painter.js
+++ b/src/js/game/buildings/painter.js
@@ -182,7 +182,7 @@ export class MetaPainterBuilding extends MetaBuilding {
                 ]);
 
                 entity.components.ItemEjector.setSlots([
-                    { pos: new Vector(1, 0), direction: enumDirection.right },
+                    { pos: new Vector(1, 0), direction: enumDirection.right, buffered: true },
                 ]);
 
                 entity.components.ItemProcessor.type = enumItemProcessorTypes.painterDouble;
--- a/src/js/game/components/item_ejector.js
+++ b/src/js/game/components/item_ejector.js
@@ -12,6 +12,7 @@ import { typeItemSingleton } from "../item_resolver";
  *    direction: enumDirection,
  *    item: BaseItem,
  *    progress: number?,
+ *    nextItem?: BaseItem,
  *    cachedDestSlot?: import("./item_acceptor").ItemAcceptorLocatedSlot,
  *    cachedBeltPath?: BeltPath,
  *    cachedTargetEntity?: Entity
@@ -38,7 +39,7 @@ export class ItemEjectorComponent extends Component {
     /**
      *
      * @param {object} param0
-     * @param {Array<{pos: Vector, direction: enumDirection }>=} param0.slots The slots to eject on
+     * @param {Array<{pos: Vector, direction: enumDirection, buffered?: boolean }>=} param0.slots The slots to eject on
      * @param {boolean=} param0.renderFloatingItems Whether to render items even if they are not connected
      */
     constructor({ slots = [], renderFloatingItems = true }) {
@@ -49,7 +50,7 @@ export class ItemEjectorComponent extends Component {
     }
 
     /**
-     * @param {Array<{pos: Vector, direction: enumDirection }>} slots The slots to eject on
+     * @param {Array<{pos: Vector, direction: enumDirection, buffered?: boolean }>} slots The slots to eject on
      */
     setSlots(slots) {
         /** @type {Array<ItemEjectorSlot>} */
@@ -61,6 +62,7 @@ export class ItemEjectorComponent extends Component {
                 direction: slot.direction,
                 item: null,
                 progress: 0,
+                nextItem: slot.buffered ? null : undefined,
                 cachedDestSlot: null,
                 cachedTargetEntity: null,
             });
@@ -97,7 +99,7 @@ export class ItemEjectorComponent extends Component {
      */
     canEjectOnSlot(slotIndex) {
         assert(slotIndex >= 0 && slotIndex < this.slots.length, "Invalid ejector slot: " + slotIndex);
-        return !this.slots[slotIndex].item;
+        return !this.slots[slotIndex].item || this.slots[slotIndex].nextItem === null;
     }
 
     /**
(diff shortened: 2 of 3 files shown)


tobspr:
So, while I like the fix to the underlying problem, I don't like the usage of null vs undefined to distinguish there. Maybe we could find a better solution?

edorfaus:
Sure. Would it be better if I used `false` instead of `undefined`? (I actually did that first, but it was less complicated to define the type comments with `undefined`.) Or some other sentinel value? (Could be almost anything, though some values will be easier to use than others.)

Alternatively, I can easily add another field on the slot that says whether or not it allows a second item (e.g. reusing the `buffered` name, matching the way it's declared), and always initialize `nextItem` to `null`. The main reason I didn't already do it that way was because I wanted to limit how much more memory usage I was adding (each new field adds to it), and a big game has a lot of these objects in it, but that could easily be premature optimization on my part.

edorfaus:
OK, I guessed you'd probably prefer the second alternative, so I implemented that - it now has a separate `buffered` boolean that defines whether or not the slot has room for a second item, with `nextItem` always being either `null` or an actual item.

tobspr:
I don't think we need this on the ejector component, since a lot of buildings have this component. Instead we should fix the item processor to buffer the items better, not?

edorfaus:
Well, if you say so. Though a lot of buildings have the item processor too, and it seemed to me that it was only buffering output by passing it off to the ejector - just that in this case, one ejector had to handle two simultaneous output items.

Looking closer at the item processor system, though, I suppose the loaded charges are a way of buffering items. (I probably understand how the game works better now, than I did the last time I looked at it.)

Hm, allowing the double painter an extra charge wouldn't help, since only the first charge is actually processed...

The idea of the charges is that the building can start processing the next charge immediately after finishing the first one, so that it doesn't lose any processing time, right? To keep up the advertised speed. (Same reason it has the bonus time thing.)

... wait what? I just noticed something odd about how charges and remaining/bonus time is handled.

As far as I can tell, it'll work fine if one charge finishes before the next comes in. But if the next comes in before the first is finished, then the bonus time from the first doesn't get applied to that next one, but to the one after that when it comes in.

I suppose this probably averages out over time to be just about right, though.

It also doesn't seem to clear out the bonus time if a new charge isn't available on the next tick. Even if it takes minutes for the next charge to be available it still gets the bonus time, it seems. I guess that's not much of a problem though, at worst it just makes the processing go slightly faster than it should.

Hm. Going back to the original problem, would this solution make sense to you? Once a charge is finished (its remainingTime hits 0), the system immediately starts processing the next charge (decreasing its remainingTime), without waiting for the first to finish ejecting. It would still eject in order, and remove charges when ejected (so new ones can come in), but wouldn't be dependent on the ejector's item buffer to avoid losing time.

I haven't really completely thought that through, so there might be issues I'm unaware of, but I'd be willing to try to implement that as a replacement for this PR. (I think the solution I'm thinking of might even end up fixing the bit about which charge the bonus time gets applied to, too.)

tobspr:
> The idea of the charges is that the building can start processing the next charge immediately after finishing the first one, so that it doesn't lose any processing time, right? To keep up the advertised speed. (Same reason it has the bonus time thing.)

Yeah

> As far as I can tell, it'll work fine if one charge finishes before the next comes in. But if the next comes in before the first is finished, then the bonus time from the first doesn't get applied to that next one, but to the one after that when it comes in.
>
> I suppose this probably averages out over time to be just about right, though.

That seems weird, but could be, are you sure on that?

>It also doesn't seem to clear out the bonus time if a new charge isn't available on the next tick. Even if it takes minutes for the next charge to be available it still gets the bonus time, it seems. I guess that's not much of a problem though, at worst it just makes the processing go slightly faster than it should.

Ah, yeah that shouldn't be a big issue, but would of course not hurt to get fixed


> Hm. Going back to the original problem, would this solution make sense to you?

Yeah that makes sense, I'd happily accept a PR regarding that! :)

edorfaus:
> That seems weird, but could be, are you sure on that?

Fairly sure, if I'm following the code flow right - I think the problem case would go like this:
1. ongoingCharges is empty, but inputs are ready; canProcess() returns true, so startNewCharge() adds a charge.
2. ongoingCharges has one item, which has its remainingTime decreased (but does not hit 0 yet).
3. inputs are ready again; ongoingCharges.length < 2, canProcess() returns true, so startNewCharge() adds a charge. bonusTime is still 0, so does not affect this charge. ongoingCharges now has two items.
4. the first charge's remainingTime goes past 0, so bonusTime is set, the output item is ejected, and the charge is removed from ongoingCharges. ongoingCharges now has one item (the one from the previous step).
5. the remainingTime of this charge gets decreased with time, but not by bonusTime, since that only happens in startNewCharge().
6. the inputs are ready again, and startNewCharge() adds a new charge, to which it applies the bonusTime from the previous charge, ignoring any that will come from the current charge.
7. loop from 4

If we assume that the processing always takes the same amount of time, and generates the same amount of bonusTime, then this would only slow things down for the second charge, since every charge after that gets the same amount of bonusTime applied to it anyway. However, if the time varies (maybe due to fps?), this might cause odd effects with speed that seems to fluctuate.

... Come to think of it, that might (at least partially) explain some of the speed fluctuations I've seen in the game, where it sometimes even goes above what the max should be (just for a moment). I've been assuming it's just due to how the speed gets measured. (Which is probably still the more likely explanation...)

tobspr:
I'll close this in favour of #1023

PR #999, opened 2020-11-22
Fix belt throughput bug

Keterr:
This is a fix for this bug reported on Discord that has been bothering some people:
image

The cause of the bug is the code for the "compressed items" optimization for belt paths. This is why the bug only appears when the belt gets "backed up" and then unblocked, because when a belt gets backed up it considers most items as "compressed" on the belt.

When the belt path offloads the first item of a compressed stack, the excess velocity that should be transferred to the next item actually gets thrown away because the rest of the compressed stack is skipped and no other candidates are found to transfer the excess velocity. Due to this there is a blockage for one tick and all items are compressed again, repeating the cycle and causing belt lines to become permanently slightly inefficient. By only skipping compressed items if an item _wasn't_ thrown away we can ensure all of the velocity is properly carried over all the time.

### DIFF
--- a/src/js/game/belt_path.js
+++ b/src/js/game/belt_path.js
@@ -1087,29 +1087,26 @@ export class BeltPath extends BasicSerializableObject {
 
             // If the last item can be ejected, eject it and reduce the spacing, because otherwise
             // we lose velocity
-            if (isFirstItemProcessed && nextDistanceAndItem[_nextDistance] < 1e-7) {
-                // Store how much velocity we "lost" because we bumped the item to the end of the
-                // belt but couldn't move it any farther. We need this to tell the item acceptor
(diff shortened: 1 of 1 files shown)


tobspr:
I don't like this fix since it causes a huge performance loss: At big belt speeds the belt is ejecting almost every frame.

I fixed the issue by instead looking at the item behind the ejected item, and check if it would get skipped: If so, I manually advance it.

Check out the 8e2efb0 commit above, would be nice to get your feedback! @Keterr  and @EmeraldBlock

Keterr:
Ahhh, I missed that my fix was skipping the compressed items optimization completely, that's definitely not ideal. 8e2efb0 looks good to me!

PR #1000, opened 2020-11-22
Visual improvement to ejector slots

Keterr:
Currently if a lane is backed up and an ejector slot is waiting to output, the waiting item often shows overlapped with items on the belt, like this:
image

Coming from Factorio and other belt-factory games this is a bit unsatisfying, it looks like there's clutter on the belts or like something is backed up.

Belt paths keep a spacingToFirstItem property that tells us how much the waiting item can peek out. So we can use this to completely avoid clutter: 
image

### DIFF
--- a/src/js/game/systems/item_ejector.js
+++ b/src/js/game/systems/item_ejector.js
@@ -159,9 +159,14 @@ export class ItemEjectorSystem extends GameSystemWithFilter {
                     continue;
                 }
 
+                const maxProgress = Math.min(
+                    1.0,
+                    sourceSlot.cachedBeltPath ? sourceSlot.cachedBeltPath.spacingToFirstItem * 1.45 : 1.0
+                );
+
                 // Advance items on the slot
                 sourceSlot.progress = Math.min(
-                    1,
+                    maxProgress,
                     sourceSlot.progress +
                         progressGrowth *
                             this.root.hubGoals.getBeltBaseSpeed() *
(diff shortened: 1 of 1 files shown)


src/js/game/systems/item_ejector.js
```diff
@@ -159,21 +159,26 @@ export class ItemEjectorSystem extends GameSystemWithFilter {
...
                 );
 
                 if (G_IS_DEV && globalConfig.debug.disableEjectorProcessing) {
-                    sourceSlot.progress = 1.0;
+                    sourceSlot.progress = maxProgress;
                 }
 
                 // Check if we are still in the process of ejecting, can't proceed then
-                if (sourceSlot.progress < 1.0) {
+                if (sourceSlot.progress < maxProgress) {
```
edorfaus, line 181: Shouldn't this still use 1.0 instead of `maxProgress`?

If I understand correctly, `maxProgress` just says how far the ejection process can get, it doesn't change when that process should be considered complete, and this check is meant to ensure that the process is complete before we hand the item off to the belt (or next building).

I think it will still work as-is, since `tryAcceptItem` below will probably return `false` (the belt not being ready for new items), but it seems like we shouldn't be trying to do that yet, and it will probably also be slower due to the extra processing.

src/js/game/systems/item_ejector.js
```diff
@@ -159,17 +159,28 @@ export class ItemEjectorSystem extends GameSystemWithFilter {
...
                 }
 
+                let maxProgress = 1.0;
+
+                // Check if there's an item blocking the ejector
+                const destPath = sourceSlot.cachedBeltPath;
+                if (destPath) {
+                    maxProgress = Math.min(
+                        1.0,
+                        sourceSlot.cachedBeltPath.spacingToFirstItem / globalConfig.itemSpacingOnBelts
```
edorfaus, line 169: This is nitpicking, but maybe use `destPath` here too, just for clarity?

Keterr: not nit-picky at all, I totally just missed that lol

tobspr:
I don't like this since this can mess up with the throughput.

Since this is a visual thing only, I decided to instead check this in the draw method and simply limit the progress there.

This also has the advantage that its more performant

tobspr:
(Will be in 1.2.1, I hope you don't mind if I use your screenshots there)

PR #1012, opened 2020-12-06
Update base-it.yaml

clemente3905:
Fixed unique item that was not translated in italian.


tobspr:
It's already translated by now, if you don't like the translation feel free to make a new PR then! :)

Issue #596, opened 2020-08-26
[Feature Request] - Better Markers


AEAEAEAE4343:
Markers currently support a shape icon or a text label. I think there should be an option to have both. e.g:
![shapez marker icon](https://user-images.githubusercontent.com/34800109/91359025-18109d80-e7f4-11ea-811f-4a44e6d4b828.png)

PyMaster22:
Good idea.

EmeraldBlock:
Duplicate of #559

PR #1014, opened 2020-12-06
#596 Better Markers

svorcmar:
Implements the feature request #596.

The shape short key may be a prefix of the label, followed by a space and the rest of the text. In that case, both the shape and the text are rendered both in the world map and the waypoint list.

Partial screenshot


EmeraldBlock:
Uh, I have a similar PR, #940

svorcmar:
Dammit, nevermind :)

PR #1017, opened 2020-12-07
enable debug env

malte9799:


EmeraldBlock:
Is this an accidental pull request?

Also, just setting that one if statement to true won't do anything - all it does is log whether or not it's a debug environment (and setting it to always true defeats its purpose). You have to change the variables cause that if statement to be true.

PR #1031, opened 2020-12-20
Error entity_manager.js ->removeDynamicComponent line 126

DJ1TJOO:
Fixed the error on line 126. 
Was: 
`        const componentId = /** @type {typeof Component} */ (component.constructor).getId();`
Is:
`        const componentId = component.getId();`

In the function component should be a type of component and not an instance.

### DIFF
--- a/src/js/game/entity_manager.js
+++ b/src/js/game/entity_manager.js
@@ -123,7 +123,7 @@ export class EntityManager extends BasicSerializableObject {
      */
     removeDynamicComponent(entity, component) {
         entity.removeComponent(component, true);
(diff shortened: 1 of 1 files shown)


tobspr:
The idea is to call it like `removeDynamicComponent(StaticMapEntity)` and not with the component instance. There could be more documentation regarding this

DJ1TJOO:
(component.constructor) gives an error because the input is already a class and not an instance. So I changed it to component.getId()

DJ1TJOO:
And in the entity class in the removeComponent function component.getId() is also used and not (component.constructor).getId()

PR #1035, opened 2020-12-24
Fix building preview I/O arrows being affected by wires

### LINKED ISSUE #648: [Bug][Alpha] Buildings on building level show outputs obstructed by wires
When placing a building in a place where wires or other wire-level components that do not interfere with buildings, the building will show inputs and outputs to be obstructed. An example is shown below with a storage-type trash and wires.

Version: Alpha 1.2.0
Environment: Google Chrome, Windows 10

image
image

EmeraldBlock:
When checking whether a building preview's inputs/outputs are obstructed, buildings in both layers are checked. This means that if there is a wires building next to an input/output, the game wrongly thinks that the input/output is blocked.

This is fixed by only checking the buildings in the same layer as the building preview.
<details><summary>Example of bug</summary>

image
</details><details><summary>Fixed version</summary>

image
</details>

Fixes #648

### DIFF
--- a/src/js/game/hud/parts/building_placer.js
+++ b/src/js/game/hud/parts/building_placer.js
@@ -510,12 +510,15 @@ export class HUDBuildingPlacer extends HUDBuildingPlacerLogic {
                 let isBlocked = false;
                 let isConnected = false;
 
-                // Find all entities which are on that tile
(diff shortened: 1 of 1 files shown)


Sense101:
is this still not fixed? it looks fine to me from what I can see

PR #1036, opened 2020-12-24
Replace gulp-sass with gulp-dart-sass

dengr1065:
Since Gulp Sass depends on Node Sass 4, which does not support Node.js 14+, it was really hard to build on Windows, because Node Sass 4 also depends on old version of node-gyp, which incorrectly detects Visual Studio version, and on other operating systems building Node Sass requires Python 2 (it reached EOL, must not be used anymore).

Node Sass 5 is already released, but LibSass is deprecated in favor of [Dart Sass](https://sass-lang.com/dart-sass). There's `sass` package on NPM, which provides ready-to-use Dart Sass, it is almost fully compatible with Node Sass. I used `gulp-dart-sass` to replace `gulp-sass` in CSS tasks, and it works perfectly. Though I had to modify animations.scss, because it contained invalid SCSS code (usage of `&` in root level).

Note: You are still using Gulp the wrong way - you should never use `gulp.src` for just one file.


tobspr:
Nice, thanks!

Issue #1039, opened 2020-12-26
Add support for gamepad


DocX:
Hi. Would you consider adding support for gamepad? Thanks!

PR #1040, opened 2020-12-26
Add basic gamepad keybindings

DocX:
Solving #1039 

Adds basic support for gamepad buttons available as keybindings

This is pretty rudimental implementation.

I think ideally the gamepad controls would be fixed (i.e. not configurable). Also it would be ideal to add sort of "tile cursor", and make use of dual joystick gamepad controls - one to move view, one to move cursor.

Needed to add keybindings for previously hardcoded mouse buttons for building placement and delete.

Some keybindings are useful to be overloaded - like delete and rotate with single gamepad button, as those actions are mutually exclusive based on the context. Given that gamepad has limited number of buttons, this is must do.

This PR is just proof-of-concept. Not all game aspects are supported yet (wires, don't know what else is there)

Mouse drag gestures may be tricky (i.e. belt placing). Not supported right now.

Does not recognize joysticks (yet)

## My mappings

* Move view - Gamepad arrows cross
* Zoom in / Zoom out - LT / RT
* Cycle buildings - RB
* Cycle alternatives - LB
* Place building - A
* Rotate building / Delete building - B

### DIFF
--- a/src/css/animations.scss
+++ b/src/css/animations.scss
@@ -1,4 +1,4 @@
-@each $num in ("changeAnimEven", "changeAnimOdd") {
+@each $animName in ("changeAnimEven", "changeAnimOdd") {
     @keyframes #{$animName} {
         0% {
             transform: scale(1, 1);
--- a/src/js/application.js
+++ b/src/js/application.js
@@ -366,6 +366,8 @@ export class Application {
 
         const time = performance.now();
 
+        this.inputMgr.processGamepadInputs();
+
         // Periodically check for resizes, this is expensive (takes 2-3ms so only do it once in a while!)
         if (!this.lastResizeCheck || time - this.lastResizeCheck > 1000) {
             this.checkResize();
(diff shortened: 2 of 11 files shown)


tobspr:
Hey, looks cool but I don't have a gamepad to test it, I'm considering it though!

Could you in the meantime cleanup the PR, so that it doesn't contain a reformatted `base-en.yaml`?

DocX:
@tobspr Hi, thank you for reviewing. Do you think this is good design anyway? I think it would be great to add support for the joysticks for the moving view, instead of relying on keybindings. But that would need some specific hardcoded implementation.

I can work on improving it over in next PR if you want. Or do you think it would be better to make the support more polished before merging?

tobspr:
I think it'd be nice to have one combined PR, so people with a gamepad can already check out if the support works :)

DocX:
Closing the PR as it's not ready yet. May continue work on it later

PR #1044, opened 2020-12-31
Apply Dark Theme Color-invert to Main Menu Top Buttons

fractile81:
Pulled the language button's color out of border and into a local variable so that can be inverted instead of the whole button.  Applied "invert" filter for other top buttons.


tobspr:
Nice, will have a look soon!

dengr1065:
This change seems to work correctly:
Light Theme
Dark Theme
The scope of this change is Main Menu state, so there should be no regressions.

PR #1048, opened 2021-01-02
Fixed the bug that caused the double painter to run slower than shown

Sense101:
The double painter outputs two shapes to the same output, and this means that the next charge doesn't start processing until the second item has started ejecting. 
To stop this, I have added a new variable in the item processor which keeps track of the extra time it takes for a building to push all the shapes to the outputs.

When starting a new charge, if the extraOutputTime is less than the time it takes for one item to travel along a belt (this is to prevent backed up buildings to start new charges with no wait time) the time it would take to process that charge is reduced by the extra time it took to output the last charge.

Hopefully the code isn't too messy :).

### DIFF
--- a/src/js/game/components/item_processor.js
+++ b/src/js/game/components/item_processor.js
@@ -96,7 +96,7 @@ export class ItemProcessorComponent extends Component {
          * @type {number}
          */
         this.bonusTime = 0;
-        
+
         /**
          * The extra time it took to empty the last charge to the outputs
          * @type {number}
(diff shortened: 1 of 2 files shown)


EmeraldBlock:
Doesn't #1023 already do this?

Sense101:
I think it does but I am doing it differently I think

Sense101:
yeah his is what I did at first but it is a lot more inefficient, running a for loop for each charge instead of just changing the actual charge TIME

Sense101:
I don't know how to fix the linting errors

Sense101:
The fix in #1023 is much better than mine, closing this pr

PR #1052, opened 2021-01-08
Added 2 @ts-ignore to building_place.js

TcePrepK:
This always gives error while modding and it is a little bit annoying.


EmeraldBlock:
Would an explicit conversion not work? Or would that go against shapez code style

Issue #522, opened 2020-08-01
Any interest in help migrating to TypeScript?


StephenFluin:
Love the game. Bought it on steam, would have paid more now that I know it's open source.

Are you interested in moving to TypeScript?

tobspr:
Thanks! You can always get it on itch.io for a price you choose yourself :)

It's been a tough decision Between TS and Js but for now I am pretty happy with the combination of JS + jsdoc + tslint. It gives me all advantages without having to transpile it. But feel free to change my mind! :)

StephenFluin:
Using what you know is definitely best. It looks like you are already using JSDoc typings though, which are much less readable, and much less strict, leading to more bugs (at least I make bugs like this). I have found that the transpilation almost never gets in the way once you have it setup.

I just compare the following:

```
    /**
     * Returns if this acceptor can accept a new item at slot N
     * @param {number} slotIndex
     * @param {BaseItem=} item
     */
    canAcceptItem(slotIndex, item) {
        const slot = this.slots[slotIndex];
        return this.filterMatches(slot.filter, item);
    }
```

and the equivalent TypeScript

```
    /**
     * Returns if this acceptor can accept a new item at slot N
     */
    canAcceptItem(slotIndex: number, item: BaseItem) {
        const slot = this.slots[slotIndex];
        return this.filterMatches(slot.filter, item);
    }
```

dengr1065:
I'd really like to see that. TypeScript has lots of amazing features other than being strict-typed language, and code would be way more clean.

tobspr:
Hm, well I think now its too late - maybe for the next project! :)

Yona-Appletree:
As someone who has migrated large projects to TypeScript, it's never too late. One of the of major benefits of TypeScript is how it can be done incrementally, and really, it's just not that hard to add typings assuming your code is fairly clean (which this is). I totally agree that TypeScript would make this project easier to read and maintain -- but I'm a TS dev by day and am no doubt biased. Great work either way.

akupiec:
I unfortunately agree with @Yona-Appletree 
Similar to him Im working on daily basic using TS (it's been already around 2 years), but I use it only in some my projects when I see benefits exactly due to transpilation time (I'm build time freek 100ms is to long when it's unnecessary ^_^ )

Im not not telling that TS it gold solution for all world problems but in some cases it's good to have types.

[...]

PR #1058, opened 2021-01-17
typescript poc in root

akupiec:
Related to #522 

TODO (if interested)
1. remove eslint checking from *.ts files
2. recheck tslint checking if working as expected


tobspr:
Just as a warning, converting the codebase to typescript is very, very unlikely to happen since it causes so many changes

akupiec:
90% of changes is done due to imports of root.ts, if I would use typescript in some new file changes to old code would be basically no existing

tobspr:
But that would be quite inconsistent then, not? If some parts of the codebase are in typescript, some not? (Don't get me wrong, I use typescript myself).

akupiec:
I see few possibilities how to approach migration to keep code consistency:

1. implement TS only for new files, and migrate old stuff on the go (for example on some significant changes) - should keep code "relatively" clean in PR's, and gradually migrate majority of code to TS (https://www.typescriptlang.org/docs/handbook/migrating-from-javascript.html)

2. (I don't think is good idea any way) rename all files from *.js to *.ts and keep changes to minimum just to satisfy TS transpiration - will keep future PR's clean but will require significant work initially.

3. (this is out of question even for me) migrate whole thing in one go ;]

PR #1080, opened 2021-03-05
Update base-cz.yaml

RevosCZ:
Decided to solve the issue with the picture overlapping the name of buildings by shortening the words into shortcuts.


tobspr:
Thanks!

PR #1086, opened 2021-03-08
Update base-tr.yaml

hcanergun:
Some "i" letters in the translation were capitalized
+  there were one english word fixed that tho


tobspr:
Ty!

PR #1088, opened 2021-03-10
Update Japanese translation of tips

Rijicho:
I fixed some strange/duplicated/ambiguous Japanese texts in tips.
Please note that the text "With modules, space is merely a perception; a concern for mortal men." may not have been translated properly as I'm not sure what it means (English L1248 to Japanese L933).


tobspr:
Ty!

PR #1090, opened 2021-03-10
Update Japanese translation

Rijicho:
Fixed some spelling inconsistencies, typos and unnatural expressions in Japanese translation file.
Also, the spaces after the exclamation marks are unified to full-width, which is a custom of Japanese writing.


tobspr:
Ty!

PR #1094, opened 2021-03-12
Prettier

DJ1TJOO:
Ran prettier on src, gulp and electron folder


DJ1TJOO:
Tobspr already did him self

PR #1096, opened 2021-03-13
Revert some inappropriate updates on Japanese tips

Rijicho:
Some Japanese translations of tips were updated by the commit 65d4c64, but it was just deletion and English duplication of some existing tips, so I fixed them in accordance with the latest English tips.


tobspr:
Nice, thank you!

PR #1097, opened 2021-03-13
Update Japanese translation

Rijicho:
I checked the all texts and updated some of them:
- Translated the added texts.
- Renamed Color Mixer from "混合機" to "混色機" because the former has only represented "Mixer".
- Fixed some wrong or incomprehensible texts (e.g. the information about tick-rate setting). 
- Fixed some inconsistency and redundancy.


tobspr:
Ty!

PR #1098, opened 2021-03-13
Update base-fi.yaml

turbomursu:
Translated the new content in lines 632-637 and 955-972. Also fixed some awkward translations to more fitting ones from before, now that I've played the game more


tobspr:
Ty!

PR #1101, opened 2021-03-13
Added notification log

DJ1TJOO:
### Adds a notification log
notifications


tobspr:
Nice feature, but currently not planned

tobspr:
Not planned but could be a mod

PR #1104, opened 2021-03-14
Enabled paused gamespeed

DJ1TJOO:
### Enabled paused gamespeed
This PR enables the paused gamespeed so you can build, but all processes are paused. It's bound to F7.

### DIFF
--- a/src/js/game/time/game_time.js
+++ b/src/js/game/time/game_time.js
@@ -9,6 +9,7 @@ import { PausedGameSpeed } from "./paused_game_speed";
 import { gGameSpeedRegistry } from "../../core/global_registries";
 import { globalConfig } from "../../core/config";
 import { createLogger } from "../../core/logging";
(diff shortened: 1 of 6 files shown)


src/js/game/time/game_time.js
```diff
@@ -185,6 +190,19 @@ export class GameTime extends BasicSerializableObject {
         this.speed = speed;
     }
 
+    togglePause() {
+        if (this.getIsPaused()) {
```
tobspr, line 194: I wouldn't do that many checks here, just do `this.setSpeed(new RegularGameSpeed())`

tobspr:
This would be nice as a mod

src/js/game/time/game_time.js
DJ1TJOO: I made it like this so there can be other gamespeeds then just regular speed

DJ1TJOO:
Adding it as a mod when modloader is done

PR #1106, opened 2021-03-14
Changed electron fs-job to invoke, for simplefying storage

DJ1TJOO:
### Changed electron fs-job to invoke, for simplefying storage
Changed the fs-job to a promised based invoke, so no need for fs-reponse to be called later.
(And it helps for the modloader)

### DIFF
--- a/electron/index.js
+++ b/electron/index.js
@@ -1,9 +1,8 @@
 /* eslint-disable quotes,no-undef */
 
-const { app, BrowserWindow, Menu, MenuItem, session } = require("electron");
+const { app, BrowserWindow, Menu, MenuItem, ipcMain, shell } = require("electron");
 const path = require("path");
 const url = require("url");
-const { ipcMain, shell } = require("electron");
 const fs = require("fs");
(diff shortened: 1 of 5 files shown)


electron/index.js
```diff
@@ -287,10 +286,14 @@ async function performFsJob(job) {
...
 }
 
-ipcMain.on("fs-job", async (event, arg) => {
-    const result = await performFsJob(arg);
-    event.reply("fs-response", { id: arg.id, result });
-});
+ipcMain.handle(
+    "fs-job",
+    (event, arg) =>
+        new Promise(async (resolve, reject) => {
```
tobspr, line 292: An async method is a promise automatically, so why not do `(event, arg) => performFsJob(arg)` ?

tobspr (changes requested):

electron/index.js
DJ1TJOO: Done that.

tobspr:
Ty! Will try to get it into the 1.3.1 release

PR #1108, opened 2021-03-14
Update base-pt-BR.yaml

LincolnVS:
Some changes in the use of the word "hotkey" and translated new section. It's fully translated.


tobspr:
ty!

PR #1130, opened 2021-04-03
Update base-en.yaml

WafflesDevAlt:
Changed the description of both tunnel tiers, as well as changed the description of belts/conveyors

### DIFF
--- a/translations/base-en.yaml
+++ b/translations/base-en.yaml
@@ -503,7 +503,7 @@ buildings:
     belt:
         default:
             name: &belt Conveyor Belt
-            description: Transports items, hold and drag to place multiple.
+            description: Transports items, click and drag to place multiple.
 
     # Internal name for the Extractor
     miner:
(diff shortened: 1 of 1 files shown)


EmeraldBlock:
I believe that HTML tags have to be preserved, due to how translations are synced

translations/base-en.yaml
```diff
@@ -503,7 +503,7 @@ buildings:
     belt:
         default:
             name: &belt Conveyor Belt
-            description: Transports items, hold and drag to place multiple.
+            description: Transports items, click and drag to place multiple.
```
EmeraldBlock, line 506: Why not change the "hold and drag" on line 1150 as well?
https://github.com/tobspr/shapez.io/blob/fa8857c6242a5a9958f59c4345c982f47613cb5b/translations/base-en.yaml#L1150

translations/base-en.yaml
```diff
@@ -519,11 +519,11 @@ buildings:
     underground_belt:
         default:
             name: &underground_belt Tunnel
-            description: Allows you to tunnel resources under buildings and belts.
+            description: Allows you to tunnel resources up to <strong>5</strong> blocks away under buildings and belts.
```
EmeraldBlock, line 522: Numbers (just numbers) are not bolded anywhere else in the translation file; I can see why bolding would make sense, but what about lines 714 and 795?
https://github.com/tobspr/shapez.io/blob/fa8857c6242a5a9958f59c4345c982f47613cb5b/translations/base-en.yaml#L714
https://github.com/tobspr/shapez.io/blob/fa8857c6242a5a9958f59c4345c982f47613cb5b/translations/base-en.yaml#L795

tobspr:
Closing because the range of tunnels is shown in the UI already

PR #1137, opened 2021-04-15
Improve the belt reader and add slight rounding

Sense101:
I have updated the belt reader system and made two changes to the code:

One - if there is no item flowing through the reader the shape output will reset to null. This makes much more sense than how it was before.

Two - the code now rounds up any value with a decimal of more than 0.8. I know this is a little bit hacky, but it perfectly fixes the problem of full belts showing as 0.1 less than they actually are, and, with a flow reader, is a perfectly normal way of making things look better in my opinion.


Sense101:
I am closing this and will create a new one at some point with no errors and cleaner code. This was not planned out properly and I think some code for this to work is not in other scripts either.

PR #1138, opened 2021-04-16
Fixes for the belt reader

Sense101:
The belt reader has a couple of problems right now:

- It can only calculate the speed if two items passed it within ten seconds of each other
- The rounding is sometimes a little low.
- The wire outputs are not in sync with the number on the reader and the shape output doesn't go away when no shapes are flowing through the reader

After going through a number of different ways of resolving this, I have done the following:
- There is now an extra variable on the belt reader component which stores the time of the last item **removed** from the list. The code now uses his variable when there is only one item in the list to correctly calculate the belt speed. 
What this means is that the belt reader can detect much lower belt speeds, up to 0.1, without actually going blank

- To fix the rounding errors I have simply added 0.01 to the throughput after its calculated and locked the throughput from going above the current max items/sec on a belt.
I have done extensive testing on this, the belt speed should now always show correctly.

- I have simply removed a lot of unnecessary code, synced the boolean output with whether or not there is any flow on the belt, and synced the shape output to it as well. This means, that even if the boolean output shows no flow, the shape output may detect that a shape is on the belt, as it shows the shape even if the belt is not flowing. You were unable to detect this before at all.


tobspr:
Looks cool, will check!

PR #1148, opened 2021-04-19
Fix formatItemsPerSecond with double = true

TimerErTim:
I noticed that when you use the double painter, the speed info says a certain speed suffixed with `(x2)`, since it processes two items at once. However, with a speed of 1.0, that suffix is lost:
Screenshot_20210419_230103

I figured I'd just create a PR, because it's only a tiny little change.


tobspr:
How does this PR fix it? Are you sure the braces are on the right positoin?

tobspr:
Thanks!

PR #1158, opened 2021-05-01
[puzzle] Prevent pipette cheats

dengr1065:
This PR makes it impossible to select forbidden buildings in puzzle mode (like extractor or constant producer) by making it obey `hasResources()` and `isBuildingExcluded()`.


tobspr:
Thanks!

PR #1159, opened 2021-05-02
[puzzle] Better trim for Puzzle Mode

dengr1065:
This PR makes trim button find actually used zone with `Array#reduce` and `Rectangle#union`, then grab the center, and align all buildings by removing all of them first, and later re-adding them after modifying position. The check for invalid building positions (outside the zone) after trimming is left for sanity. It also makes sure the zone is at least 2x2 (from globalConfig).


tobspr:
Already implemented, but still thanks!

PR #1162, opened 2021-05-04
Update base-es.yaml plsss!

AlvarGD:
HUB -> Centro 
HUD -> Interfaz (almost no-one that knows spanish uses "HUD")
All (hopefully) things that weren't translated now are
Decimals and thousands are now how they should be in spanish: 1.000.000,69


tobspr:
Thanks!

PR #1163, opened 2021-05-05
Filter out shifted key presses.

garretsimpson:
Only accept key presses that do not have Shift, Ctrl or Alt pressed.

This does not alter the behavior of existing game features such as Ctrl + mouse drag to select an area.


garretsimpson:
Did not mean to make a PR for entire copy-paste branch.  Closing.

PR #1164, opened 2021-05-06
[Puzzle] Added ability to lock buildings in the puzzle editor!

Sense101:
One feature I have been thinking that was really been missing from the puzzle beta so far is the ability to lock buildings of your choice to make them unusable in play mode. 

Don't ask me why, but I have made it for you!

In edit mode:
- Standard buildings now have a small lock icon above them in the puzzle editor.
- When you press it, it locks that building from use and removes all instances of that building from the puzzle.
- The building then looks locked but instead of the locked building icon, it is still the icon for the building, which looks so much cooler!
- Pressing on the smaller icon of the locked building unlocks it, allowing you to select it again!

The puzzle serializer detects which buildings are locked and saves their classes to the puzzle data.

When you then play a puzzle it adds those buildings to the hidden buildings, meaning you cannot use them to complete the puzzle!

I haven't actually been able to test the play mode as I don't have access to the backend, but from a code perspective it should all work fine.

Much thanks to Dengr and Shrimp for helping me get this to work!


tobspr:
Wow nice, thanks!

PR #1165, opened 2021-05-07
Clipboard copy and paste

garretsimpson:
This PR modifies the behavior of copy and paste of blueprints.  This enables the following features:
- Copy and paste of blueprints between worlds.
- Import of blueprints created by external applications.
- Saving and sharing of blueprints in text form (e.g. to discord or reddit).

Copy will write a text-based serialized version of the blueprint (currently selected buildings) to the system clipboard.

Paste will attempt to read a blueprint from the system clipboard.  If that fails, it falls-back to the current behavior of using the last blueprint created.

This change is currently guarded by the development flag "useClipboard".  When it is not set, there is no change in behavior.  But this can be easily removed, or perhaps modified to use an in-game configuration instead.

This PR also includes a change to the handling of key presses.  This change ignores "shifted" key presses (Shift, Ctrl, Alt).  This does *not* alter game features that use combos such as Ctrl+MouseDrag.  This change is not needed for the copy and paste feature and should probably have been put on a separate branch.  It is contained in a single commit (the first one) and can be easily removed if it is not desired.


tobspr:
Will check!

tobspr:
Hey! I like the idea, but would like to have more validation  on the string side, as well as compressing blueprints with lz-string so they can be more easily copy / pasted on reddit/discord etc.

Also would be nice to have the input change seperated :)

garretsimpson:
Hi.  I've made the code changes you requested:
- Removed the key press changes.
- Compressed the serialized data.  I see about 5:1 compression.
- Added a few more checks to the validation.  Do you have suggestions for additional checks?

BTW - Do you think this worth adding an in-game config (ie. toggle on Advanced Settings page), instead of a debug flag?

tobspr:
I'll check and get back to you :)

mekb-turtle:
this looks good! people are definitely going to (ab)use the copy/pasting to text if this gets merged

tobspr:
Should be a mod for now I think :) M

PR #1166, opened 2021-05-07
Update base-uk.yaml

Vilser1:
I dont really know how to do "yarn syncTranslations" so...


EmeraldBlock:
I believe `yarn syncTranslations` is only necessary if you change the *format* (like the `<>` tag stuff) of translations, as it makes sure everything follows the English translation's format

tobspr:
Thanks!

PR #1170, opened 2021-05-11
Added a button to return to the menu after a puzzle is completed

Sense101:
This makes it much easier to navigate puzzles, if you want to stay on a puzzle you can, but if you want to return to the menu you don't have to press continue and then leave.


tobspr:
Cool, thanks!

PR #1185, opened 2021-06-04
Update base-cz.yaml

RevosCZ:
Translated some of the parts for the Puzzle DLC, will try to get the rest done aswell. When the dlc is out, will have to go through the translations ingame, and if something seems odd, will update the translation.


tobspr:
Cool, thanks!

PR #1195, opened 2021-06-05
Base-tr.yaml Puzzles Translations

dobidon:
- Translated Puzzles DLC Additions

Discussion:
- Unsure about Line 429 Trim Zone.


EmeraldBlock:
I believe `trimZone` means the name of the button that will shrink the borders of the puzzle area (for example, if the entire left row is empty, the left border will move closer).

dobidon:
> I believe `trimZone` means the name of the button that will shrink the borders of the puzzle area (for example, if the entire left row is empty, the left border will move closer).

Got it, thanks. I changed the translation according to it. If it does not fit after the DLC shows up, it can be changed later on.

tobspr:
Thanks!

PR #1197, opened 2021-06-06
Optimization - Use map to store input slots rather than an array

Kietyo:
While doing some performance profiling I noticed that the tryTakeItem was iterating through an array every time in order to find whether or not a slot is populated. I realized that maybe using a map would be better even though it is still constant time to iterate through the array of slots (max # of input slots if 5). I think this will matter late game when players have more than hundreds of thousands of buildings.

From doing this optimization, I also found that we were doing a lot of needless copying of items in the `startNewCharge` method since we can just directly use the map that we've switched to using.


tobspr:
Have you checked if this actually improves the performance? I'm honestly not sure that a map will actually be faster here, but I'd be happy to get proven wrong!

PR #1198, opened 2021-06-10
kor translation for Puzzle DLC

YJSoft:
I also translated missing translation for wire update.

I didn' translated `nothernlion_comment`, `notch_comment`, `steam_review_comment` since translation might change Its meaning. but If you want to, I can translate that too.

p.s.) Translation for `Short key` is `짧은 키` for now, but if anyone have better idea, I can change that too.


tobspr:
Yeah not translating the reviews is right :) Thanks a lot!

PR #1200, opened 2021-06-10
Simplify `readAsync` method

CyberGsus:
I've been working on reducing the brain power needed to understand the workings of the `readAsync` method and
make the code more cache and v8-friendly by naming callback functions.

Changes:
  - Removed unnecessary compression step
  - Moved arrow callback functions to named functions
  - added some notes

### DIFF
--- a/src/js/core/read_write_proxy.js
+++ b/src/js/core/read_write_proxy.js
@@ -161,141 +161,196 @@ export class ReadWriteProxy {
 
   // Reads the data asynchronously, fails if verify() fails
   readAsync() {
+
+    /**
+      * Decompresses raw data through `decompressX64`
+      * Fails if decompression failed or
+      * data is too small.
+      * @param {string} data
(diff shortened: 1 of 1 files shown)


tobspr:
`code more cache and v8-friendly by naming callback functions.`

Any reference for why this should be more cache friendly? 

Also, have you tested if it makes any difference in performance?

CyberGsus:
Yeah, forget the v8 thing. I don't know what I was thinking about this morning. I have yet to determine A) when the file not found error is triggered to be able to measure any difference, and B) measure the loading of save files (existent) vs the current way.

CyberGsus:
If I don't get any better performance at loading files I will close this PR, as the point of this is to reduce game load time.

CyberGsus:
I triggered the loading of my own save file (which I must recognize it has not many buildings, but it's not related to what I was testing) on each test so I could get the functions to run.

PR #1201, opened 2021-06-10
Update base-cz.yaml

Kikismine:
Please updata my file

### DIFF
--- a/translations/base-cz.yaml
+++ b/translations/base-cz.yaml
@@ -1165,56 +1165,56 @@ tips:
     - Stisknutím F4 dvakrát zobrazíte souřadnici myši a kamery.
     - Můžete kliknout na připínáček vlevo vedle připnutého tvaru k jeho odepnutí.
 puzzleMenu:
-    play: Play
-    edit: Edit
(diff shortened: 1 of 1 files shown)


DerFeldspatz (commented):
Syntax seems to be ok.
I don't understand czech but there seem to be 5 different translations for `puzzle` in lines 1171, 1174, 1175, 1196 and 1207. Can you explain or someone who's fluent check?

RevosCZ:
Already made one pull request before, so when mine goes through, there will be conflicts... Aswell, not all, but some of these translations are missing diacritic and letters.

DerFeldspatz:
I've checked your PR and it seems to me that you consistently use `puzzle`. Might be better to keep this up. 
However, all the lines this PR features also appear in #1189 . Not to be rude but I would recommend you @Kikismine to close this PR and comment and discuss your suggestions in #1189 since it seems to be more developed... 
Feel free to ask back if you don't fully understand me ✌️

tobspr:
Can you check the conflicts and the comment from @DerFeldspatz ?

PR #1203, opened 2021-06-12
Internect typo fix

RolandSzaboDesign:
Fixed a small typo in all of the translation files.


tobspr:
Good catch, I'll fix it in a separate commit!

PR #1206, opened 2021-06-12
Dlc hungarian translation

RolandSzaboDesign:
Done with the full HUN translation for the Puzzle DLC.
I'll do some updates when the game is testable if any improvements are needed.


DerFeldspatz:
The PR as itself seems fine to me, but could you please have a look at the yaml-lint warnings for base-hu and fix them? Even if they are not caused by you, just add some line breaks for the according lines ✌️

tobspr:
Thanks!

PR #1208, opened 2021-06-14
Correct zh-TW translation in tutorial 21_1_place_quad_painter

w84miracle:
`painter` should be `上色機`
`切割機`  means `cutter` in zh-TW


tobspr:
Thanks!

PR #1210, opened 2021-06-14
[Fix] `compressInt` now uses cache properly

### LINKED ISSUE #1209: [Bug] `compressInt` can't take advantage of the cache as it always writes the `0` key.
The function [`compressInt`](https://github.com/tobspr/shapez.io/blob/1046e7d4bd9fba87fadc660902c431d85a2684fd/src/js/savegame/savegame_compressor.js#L12) will always overwrite the key `0`, because of the `do while` loop which will eventually end up in `i` being `0` (from the `Math.floor` division). This compromises the use of the cache, not only because it always overrides the key `0` when computing a value, but it also always misses the cache ,as the value of `i` when checked against the cache is > 0.

You can verify this by going into node, pasting the function definition
and then defining `let compressionCache = {}` and `const charmap = "...."`, both pasted from the source (Lines 4 and 1 respectively), then calling the function with any number (e.g `compressInt(10)`) and checking the contents of `compressionCache`. It always will be `{ '0' : /* <last computed value from compressInt> */ }`

I'll submit a PR to fix this.

CyberGsus:
This PR fixes #1209, check it for all the info!

### DIFF
--- a/src/js/savegame/savegame_compressor.js
+++ b/src/js/savegame/savegame_compressor.js
@@ -13,15 +13,20 @@ function compressInt(i) {
     // Zero value breaks
     i += 1;
 
-    if (compressionCache[i]) {
-        return compressionCache[i];
+    // save `i` as the cache key
+    // to avoid it being modified by the
+    // rest of the function.
+    const cache_key = i;
+
+    if (compressionCache[cache_key]) {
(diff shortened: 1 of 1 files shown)


tobspr:
Nice catch! Thank you

PR #1211, opened 2021-06-14
Italian puzzle translation

Doc-Ninja:
Notes for reviewers:
-Celebs comments are not translated
-Bad puzzle title has been translated with a more generic loading failed
-I left "Puzzle DLC" everywhere even if not quite correct in Italian as it is found as such on steam
-Backend error types are not translated


Doc-Ninja:
This fails yaml lint because of errors in other languages, only a single warning in the italian file

tobspr:
Cool, thank you1

PR #1225, opened 2021-06-23
Replace some punctuation marks by Chinese version and optimize format

sooxin:


tobspr:
Nice, thank you!

sooxin:
> Nice, thank you!

It's my pleasure :-)

PR #1226, opened 2021-06-23
Fix for cheating puzzles by quickly switching belts

Sense101:
This PR changes the way the goal acceptor works, simplifying a lot of the code and preventing macro exploits.

I have already tested it thoroughly and there are no problems with it at all. I actually think it looks nicer than the original version!

Unfortunately puzzles can still be cheated by quickly placing and removing cutters, but its almost impossible to fix that one.


src/js/game/components/goal_acceptor.js
```diff
@@ -31,19 +31,26 @@ export class GoalAcceptorComponent extends Component {
...
 
-    getRequiredDeliveryHistorySize() {
+    // clears items but doesn't instantly reset the progress bar
+    clearItems() {
+        this.lastDelivery = null;
+
+        this.currentDeliveredItems = 0;
+    }
+
+    getRequiredItemsPerSecond() {
```
EmeraldBlock, line 50: Shouldn't this be seconds per item?

src/js/game/components/goal_acceptor.js
```diff
@@ -31,19 +31,26 @@ export class GoalAcceptorComponent extends Component {
 
     clear() {
         // the last items we delivered
-        /** @type {{ item: BaseItem; time: number; }[]} */
-        this.deliveryHistory = [];
+        /** @type {{ item: BaseItem; time: number; }} */
+        this.lastDelivery = null;
+
+        this.currentDeliveredItems = 0;
```
EmeraldBlock, line 37: Why not use `this.clearItems()` here?

src/js/game/systems/goal_acceptor.js
```diff
@@ -24,13 +24,16 @@ export class GoalAcceptorSystem extends GameSystemWithFilter {
...
-                d =>
-                    now - d.time < globalConfig.goalAcceptorMinimumDurationSeconds && d.item === goalComp.item
-            );
+            if (!goalComp.lastDelivery) {
+                continue;
+            }
+
+            if (now - goalComp.lastDelivery.time > goalComp.getRequiredItemsPerSecond()) {
+                goalComp.lastDelivery = null;
+                goalComp.currentDeliveredItems = 0;
```
EmeraldBlock, line 33: Why not use `goalComp.clearItems()` here?

EmeraldBlock (commented):
Just some minor stuff, haven't looked at puzzle mode code before so sorry if I misinterpreted something!

src/js/game/systems/item_processor.js
```diff
@@ -572,23 +573,23 @@ export class ItemProcessorSystem extends GameSystemWithFilter {
...
         const now = this.root.time.now();
 
+        goalComp.currentDeliveredItems = Math.min(
+            goalComp.currentDeliveredItems + 1,
+            globalConfig.goalAcceptorItemsRequired
+        );
+
+        if (goalComp.item && !item.equals(goalComp.item)) {
+            goalComp.clearItems();
+        }
```
EmeraldBlock, line 583: Why not put the first chunk in an `else` after the second chunk?

EmeraldBlock (commented):
Just one more thing lol (also, this PR would make the acceptor percentage indicator less jittery, that's nice)

src/js/game/components/goal_acceptor.js
Sense101: You are correct, it should be. ooops

src/js/game/components/goal_acceptor.js
Sense101: Because that also clears the animation completely

src/js/game/systems/item_processor.js
Sense101: yeah will do, thanks :)

src/js/game/components/goal_acceptor.js
EmeraldBlock: but isn't `clear()` the one that clears the animation?

tobspr:
Nice looks good! Testing it now

PR #1229, opened 2021-06-24
Added a button to clear all non fixed buildings in both modes

Sense101:
Having this button is really useful for playtesting, and also for when you want to just clear your solution in the play mode. I'm not an expert at css, but I think it turned out ok :)


tobspr:
Nice! I'll adjust the check a bit but looks good otherwise

PR #1230, opened 2021-06-24
Add back copy/paste in puzzle mode.

Sense101:
This PR adds back copying and pasting for the puzzle editor and play mode.

While in the play mode, you can't select any building that you can't delete.

Tunnel weaving, anyone?


tobspr:
Nice, also testing it now!

PR #1233, opened 2021-06-24
Fixed when achievements are given

DJ1TJOO:
With this pr achievements are given on the correct level

Co-authored-by: TcePrepK <TcePrepK@users.noreply.github.com>

PR #1234, opened 2021-06-24
This little change should achievement bugs

TcePrepK:
Level achievements are bugged for a lot time recently and for some reason noone made pr to fix this so yeah... Anyways this bug causes player to get level achievements before getting to that level. So removing "=" from there "should" fix these bugs.


gsquaredxc:
See #1233, published right before you.

tobspr:
Done with #1233

PR #1238, opened 2021-06-25
Cleanup of buildings code and rename to reset puzzle

Sense101:
The code for the clear buildings code was still a bit messy, mainly because hen I made the PR I didn't pick the best way to go about it. Hopefully its better now

The goal acceptors now correctly clear when the puzzle is reset.

The reset button is now always called reset puzzle, I think it makes more sense anyway.


Sense101:
Partly merged, I think next time I'll stick to moderating 😄

PR #1240, opened 2021-06-25
Copy from ejectors

Sylviromi:
With this pr, "copy wire value" key will copy from ejectors instead of entities. What I am trying to say is, this pr will make copying from transistor, virtual processors etc. possible. It does this with calculating the position of sprite of value, finding the distance of it to mouse, then it just copies it if it is close enough.


EmeraldBlock:
Feel maybe instead of detecting distance to each pin you could divide the tile into sectors by the locations of the ejectors (ie, use the angle of the mouse relative to the center of the tile)? Would also mean larger clickable areas for fewer pins, I may try implementing later

PR #1241, opened 2021-06-25
Update base-de.yaml

DerFeldspatz:
I sorted, streamlined and corrected the german translation.
There also were two spelling mistakes I found in the base file.


tobspr:
Super, vielen vielen Dank!

PR #1245, opened 2021-06-27
Added a shape tooltip when holding alt

Sense101:
I was bored so I made another PR :)

You can press alt while hovering over buildings now to show what the building is producing. This should help with no belt puzzles.

There is one problem with this which I was unable to fix, that is, for some reason when moving using keyboard controls it doesn't register it as the mouse moving, so it doesn't update the position. This seems to work on other HUD parts like the mass selector, but I couldn't figure out why.

There shouldn't be any other issues, but don't quote me on that.

I really need to stop wasting time making PR's, I have better things to do 😛


tobspr:
Could you rebase your branch on the latest master? It seems its including some old changes ...

tobspr:
Also I'll probably change the code a bit, since instead I'd just read the mouse position and see whats below - I can do that too though

Sense101:
yeah I'll rebase it. I can change it to checking for the mouse position as well as I already worked out how to do it that way.

Sense101:
I'll reopen another PR with it rebased, its the easiest way to do things

PR #1246, opened 2021-06-28
Credits

Bagel03:
This PR adds a "Credits" state and gulp task, which displays all the people who have had a pull request merged into the game. This includes bug fixes, features, and translations. All PRs and Users are linked, and separated into 3 categories: `Special Thanks`, where you can hand pick who is there and what they did (Right now it is only Peppsen), `Translators`, for PR's that only edited files in the `translations` folder, and `Contributors` for all other PRs.  All sections are collapsible .

There is currently a build running at https://bagelshapez.web.app/, go to the `About This Game` page, and click the credits link. 

Some notes:
The gulp task is not integrated with most of the other build tasks, this is intentional. It has to download a lot of files, which takes a good chunk of time. Also, running it a lot of times in a row will cause it to be ratelimited by github. Run `yarn gulp contributors.build` to build it.

By default, github only gives you 60 API requests an hour. To get around this you need a Personal Access Token. This can be done by going to your settings, developer settings, personal access tokens, and generating a new token. Then paste that into the 
`personalAccessToken` variable at the top of `gulp/contributors.js`.

image
image


Galo223344:
I feel like translations should be categorized by `language:[users]` rather than `users:[commits]`

Bagel03:
> I feel like translations should be categorized by `language:[users]` rather than `users:[commits]`
Its a little hard to get exactly which language a PR adds without creating a large lookup table and counting on the fact that they only updated one file each run. Obviously the first example in that photo doesn't look too good, but most of the other uses that gave their PR's better name. What I am thinking about is replacing all names that are just "update base-en" to something like "Updated English translations", however IDK if that would fit with the rest of the page.

tobspr:
Nice that looks really cool! will check it out

PR #1247, opened 2021-06-28
Aggregate map chunks in overlay.

PFedak:
Overlay rendering performance seemed bottlenecked by drawImage calls. To
reduce both the number of calls and the number of different source
buffers, cache overlay buffers for squares of chunks. This adds a very
small extra cost for updates (one additional drawImage) and some cost
for drawing chunks outside of view, but this is more than made up for by
the savings.

By default, the aggregate are 4x4 squares of chunks.

### DIFF
--- a/src/js/game/map_chunk_aggregate.js
+++ b/src/js/game/map_chunk_aggregate.js
@@ -1,0 +1,146 @@
+import { globalConfig } from "../core/config";
+import { DrawParameters } from "../core/draw_parameters";
+import { MapChunk } from "./map_chunk";
+import { GameRoot } from "./root";
+import { drawSpriteClipped } from "../core/draw_utils";
+
+export const CHUNK_OVERLAY_RES = 3;
+
+export class MapChunkAggregate {
+    /**
+     *
+     * @param {GameRoot} root
+     * @param {number} x
+     * @param {number} y
+     */
+    constructor(root, x, y) {
+        this.root = root;
+        this.x = x;
+        this.y = y;
+
+        /**
+         * Whenever something changes, we increase this number - so we know we need to redraw
+         */
+        this.renderIteration = 0;
+        this.dirty = false;
+        /** @type {Array<boolean>} */
+        this.dirtyList = new Array(globalConfig.chunkAggregateSize ** 2).fill(true);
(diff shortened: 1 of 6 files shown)


PFedak:
Not sure if there's just something wrong with my setup, but fully zoomed out in overlay mode I was getting single digit FPS. This brings it back to max when just scrolling. I'm far from an expert on canvas, so maybe there's a way to encourage all the buffers to be more efficiently in memory, but this accomplishes something similar. 

I didn't change the map resource background logic, as it wasn't a huge impact. I'd love a better name than "aggregate", but "chunk" and "patch" were already used.

src/js/game/map_chunk_aggregate.js
```diff
@@ -0,0 +1,146 @@
...
+            dpi: 1,
+            redrawMethod: this.generateOverlayBuffer.bind(this),
+        });
+
+        const dims = globalConfig.mapChunkWorldSize * globalConfig.chunkAggregateSize;
+        const extrude = 0.05;
+
+        // Draw chunk "pixel" art
+        parameters.context.imageSmoothingEnabled = false;
+        drawSpriteClipped({
```
PFedak, line 114: Worth noting there's a subtle visual change here: the individual chunk images aren't "extruded", just the whole. I didn't try playing on "potato mode" to see if that leads to visual issues

tobspr: shouldn't be an issue :)

tobspr (approved):

tobspr:
Looks cool and has been on my todo list for *ages*! Will check out

PFedak:
Also, just making sure because I always get confused how git handles it, you're using CRLF/windows line endings in this repo, right? I was also thinking about looking into caching the blueprint preview, since it seems like that's redrawn from scratch currently.

tobspr:
I think it should be LF and then automatically get converted to CLRF when you check it out (Should be the git default setting). But not 100% sure

PFedak:
Whoops, sorry, forgot I had done this from my fork's master branch. Also, as far as line endings, I had spot checked a couple of files before. There's actually a mixture of windows- and unix-style line endings in the repo. I'll make sure I'm using unix.

PR #1248, opened 2021-06-28
Update base-de.yaml

FabulousCodingFox:
Not only Chrome is fastest, also other chromium-engine based Browsers are as fast.


EmeraldBlock:
Although this may be good advice, note that the game itself checks for Chrome and not the other browsers. This also means an inconsistency between translations - the other translations still only say Chrome.

FabulousCodingFox:
> Although this may be good advice, note that the game itself checks for Chrome and not the other browsers. This also means an inconsistency between translations - the other translations still only say Chrome.

Yeah makes sense though.

PR #1249, opened 2021-06-28
Update base-es.yaml

Galo223344:
Hi! Is there any way to recover the comments from en-base.yaml? They were very helpful but somehow got removed.


tobspr:
Thanks! Unfortunately not :/

PR #1251, opened 2021-06-30
Added shape tooltip - final version

Sense101:
You know what this is, it should be a lot cleaner now.

I'm annoyed now about my messy commits before I knew how github worked 😆

### DIFF
--- a/src/js/game/hud/parts/shape_tooltip.js
+++ b/src/js/game/hud/parts/shape_tooltip.js
@@ -1,0 +1,89 @@
+import { DrawParameters } from "../../../core/draw_parameters";
+import { Vector } from "../../../core/vector";
+import { Entity } from "../../entity";
+import { KEYMAPPINGS } from "../../key_action_mapper";
+import { BaseHUDPart } from "../base_hud_part";
+
+export class HUDShapeTooltip extends BaseHUDPart {
+    createElements(parent) {}
+
+    initialize() {
+        /** @type {Vector} */
+        this.currentTile = new Vector(0, 0);
+
+        /** @type {Entity} */
+        this.currentEntity = null;
+
+        this.isPlacingBuilding = false;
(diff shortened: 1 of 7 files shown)


src/js/game/hud/parts/shape_tooltip.js
```diff
@@ -0,0 +1,89 @@
...
+        this.root.signals.entityQueuedForDestroy.add(() => {
+            this.currentEntity = null;
+        }, this);
+
+        this.root.hud.signals.selectedPlacementBuildingChanged.add(metaBuilding => {
+            this.isPlacingBuilding = metaBuilding;
+        }, this);
+    }
+
+    update() {
```
tobspr, line 29: Why have an update() and an active() method? Just make the update() method a isActive() method and call it in the draw() method. Or am I missing something?

tobspr (changes requested):

src/js/game/hud/parts/shape_tooltip.js
tobspr: Then you also don't need to store an active and currentEntity flag

tobspr: Let me know if you want me to adjust it myself :p

Sense101: yeah that makes sense. And then update isn't even needed, it can all be in the draw method. Will do that now

Sense101: Okay that's done, I could also just put the isActive checks in draw, but it may just be simpler to leave it

Sense101:
currentEntity is still required to be cached though as far as I can see, unless you want to recheck the tile every tick

Sense101:
This should be ready to be merged now, if you plan on adding this. Let me know if there are any further changes you want

PR #1254, opened 2021-07-02
Added letters and numbers on the side of the zone.

Sense101:
This was just an idea that I had, and I don't necessarily feel that this should be merged. It's just a very simple change and it would be nice to know what you think about it.

Right now I think it looks out of place, and I think having an option in settings for this would work better, but that would require a section of the settings which only shows if you have the puzzle dlc, which is annoying to do.


EmeraldBlock:
nonsuspicious commit history

Sense101:
Marked as draft as I want to add a puzzle section in the settings and make this an option there.

PR #1257, opened 2021-07-04
Simplify item processor slot logic.

PFedak:
Track input slot assignment by array index rather than including extra
data in a new object. This greatly reduces the amount of garbage
generated by tryTakeItem and other item process operations.

### DIFF
--- a/src/js/game/systems/item_processor.js
+++ b/src/js/game/systems/item_processor.js
@@ -32,8 +32,7 @@ const MAX_QUEUED_CHARGES = 2;
  * Type of a processor implementation
  * @typedef {{
  *   entity: Entity,
- *   items: Array<{ item: BaseItem, sourceSlot: number }>,
- *   itemsBySlot: Object<number, BaseItem>,
+ *   items: Array<BaseItem?>,
  *   outItems: Array<ProducedItem>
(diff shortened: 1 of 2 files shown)


PFedak:
This goes a bit further than #1197 to improve performance in item processor logic. I see that that PR is "on hold", and I share your skepticism about the map having a large impact. In my experience eliminating unnecessary garbage generation is an important tool for improving js performance. I'm not sure if you have a standard benchmarking process, but this change took my median tick time from about 11.1 ms to 9.5 ms, just sitting idling watching the hub.

src/js/game/systems/item_processor.js
```diff
@@ -262,13 +255,6 @@ export class ItemProcessorSystem extends GameSystemWithFilter {
 
         // First, take items
         const items = processorComp.inputSlots;
-        processorComp.inputSlots = [];
```
PFedak, line 265: Removing this involves assuming the item processor won't try to take any input during the handler and signal processing, which is true currently but could be a limitation on future behavior. I don't think this change in particular has a large impact on performance.

tobspr:
Thanks! Made a few adjustments, but overall great PR :)

PR #1265, opened 2021-07-14
Fix to producer

TcePrepK:
Current hiddenBuildings array doesn't allow you to have producer if you are dev on regular gamemode. But with adding this little check, we can fix it :)


EmeraldBlock:
What about sandbox mode?

PR #1266, opened 2021-07-14
Fix to producer

TcePrepK:
Current version of hiddenBuildings, removes producer from sandbox and dev build. This change should fix it


tobspr:
Fixed it, did some adjustments :)

PR #1268, opened 2021-07-22
fixed the output speed of producers being too fast sometimes

Sense101:
Simply adds a timer to output items at an interval, which is anyway more neat than trying to output every frame.


EmeraldBlock:
imo trying to eject each tick is better; it comes down to just checking if the slot is occupied. Plus, no dependence on puzzle mode and no additional time calculation worries (also what about item producer building?)

Sense101:
True, maybe this simply needs more thought. Marking as draft for now

tobspr:
Yeah I don't think this is a good idea :D

PR #1270, opened 2021-07-25
Respect "offerHints" in puzzle editor

dengr1065:
Stops Puzzle Editor game mode from adding Puzzle Editor Controls HUD if `offerHints` setting is disabled. Done with same code as already used in Regular game mode for interactive tutorial, got 5 positive votes on Discord suggestion.


tobspr:
This is actually by design. If, then it should be a different setting.

Reason: If you have disabled hints for the main game, you should still see hints the first time you play the puzzle DLC to understand how it works.

PR #1272, opened 2021-07-27
Hotfix to prevent pasting locked buildings

Sense101:
Simple change which makes sure that a building which is locked can never be placed.

I changed this to only check in the editor as it is never going to apply anywhere else anyway.

I think this may partially break my PR which adds a test mode but it needs some minor changes anyway.


tobspr (approved):

tobspr:
Thanks!

Issue #1194, opened 2021-06-05
Double tip in tip list


DirkieDurky:
I noticed the tip: "Mix all three primary colors to get white." exists twice. Is this on purpose or can it be removed?

PR #1274, opened 2021-07-31
Update base-nl.yaml

stormweb-dev:
New translation added


DerFeldspatz:
Why did you delete one of the tooltips?

DJ1TJOO:
> Why did you delete one of the tooltips?

I think they thought it was still double in the Dutch translation: https://github.com/tobspr/shapez.io/issues/1194

DerFeldspatz:
I remember the tooltip for white color being double in both english and german translation with slightly different phrasing. So I think it's intended, no?

DJ1TJOO:
> I remember the tooltip for white color being doubled in both english and german translation with slightly different phrasing. So I think it's intended, no?

It is intended, my phrasing was a little wrong

tobspr:
Thanks!

Issue #1126, opened 2021-03-26
Wires tutorial reappears if any switches are set to off


xxiii23:
The wires tutorial doesn't seem to recognize when its been completed and shouldn't appear anymore (is this the first/only switch on the entire map?); or it gets retriggered every time a switch is set to off. So instead of having switches that are set to off, one has to delete those switches instead. (Or disable tutorials, which I don't want to do).

EmeraldBlock:
Known issue, the more general "bug" is that all tutorials do not remember that they are completed and instead re-show if they are undone. A sort-of-fix is to disable tutorials in settings

PR #1280, opened 2021-08-08
Make tutorials monotonic.

PFedak:
Don't reshow tutorial steps that have been completed.

Also make video tutorials respect the tutorial setting.


PFedak:
Fixes #1126, which seems like the only case people are likely to encounter/be annoyed by. I can see some value in repeating the steps for the very first tutorials if e.g. someone accidentally disconnects their conveyor from the hub. Maybe just the completion state should be sticky.

Issue #1180, opened 2021-05-31
[QoL] Be able to see conveyer plan when in zoomed out view. 


Kietyo:
What happens to me frequently:

1. Click and drag conveyer
2. Accidentally zoom out too much
3. The conveyer track completely disappears so I have to repeat (1) again

This is quite frustrating so I would like the conveyer plan to work when in the zoomed out view.

Issue #1151, opened 2021-04-23
Level complete clears placing tool


nevjmac:
Once you've got a MAM running and clearing levels relatively quickly. Everytime the pop up comes up saying level complete it clears whatever you were in the process of building if holding shift, or whatever you have in selected to build

EmeraldBlock:
Duplicate of #998

PR #1282, opened 2021-08-09
Improve building experience involving overlay.

PFedak:
Allow continuing and placing belt/wire path plans in the map overlay
view. All logic is the same except for not drawing the arrows when
zoomed out.

Also remember previously selected building when an overlay prevents
building and switch back when appropriate.

Clearly wasn't that important, but abortDragging didn't actually change
the dragging variable. The value of this variable now matters for
properly cancelling the belt/wire planning state after placing in
overlay view.


PFedak:
This implements #1180

PFedak:
This doesn't quite work yet because of some other places that the current meta building is cleared (e.g. switching edit layer). I wonder if those would also benefit from the switching behavior (could help with #1151 too)

PR #1283, opened 2021-08-09
Trim bugfix

Sense101:
This PR fixes a minor bug which meant that trimming would remove the items on goal acceptors. It also cleans up the code to something more logical :)


tobspr:
Awesome, thanks!

PR #1286, opened 2021-08-27
Translate 'clear belt' for korean

xnuk:
- Note: I translated as like 'Cleaning/Sweeping belt'. I'm not sure if 'Emptying belt (벨트 비우기)' would be a proper word instead.
- Note 2: First I thought 'clear belts' does delete all belts only in a selected area, but actually it deletes objects on the belts only.


tobspr:
Thanks!

PR #1289, opened 2021-08-29
Fix italian translation

battistalonardi:
Added italian translation for text that were still in English. Made uniform translations for rotater and painter


tobspr:
Awesome, thank you!

PR #1295, opened 2021-09-03
Different Loading Tips For Creating and Playing Puzzles

Bagel03:
This PR does what the title says: You now get different loading screen tips depending on what you are about to do. For example, when loading a puzzle, hints about working through a puzzle show up, that would not appear in the normal game. The list is not complete, but I wanted to create the PR just to get the code out there because obviously translations would be needed, and more/better suggestions would be very nice. 

So far, when you are about to play a puzzle, the options are:
```
    - The double painter can be used to combine belts.
    - Working through puzzles one part at a time often works.
    - When starting a puzzle, find out where different buildings fit.
    - In puzzles without belts, pressing <b>ALT</b> will let you see shapes more easily.
    - Don't get worked up! Try looking at things from a different angle.
    - If a puzzle has a throughput requirement, remember to split belts in a tree-like fashion.
    - Think about what order you can create your shapes in.
    - Don't get scared by big puzzles, try working through them before giving up.
    - Play the tutorial puzzles to find some useful and easy-to-use tricks.
    - Remember, staring at a puzzle won't help! Place some blocks and see where you end up.
    - Don't forget to like a puzzle if you enjoy it.
    - If you run into a puzzle you know is 100% not possible, make sure to report it.
    - Been playing a lot of puzzles? If so try creating your own for others to enjoy.
    - A unwise man once said "It's been so long, this puzzle is not possible".
    - Remember, it is not a speedrun. Take your time and solve the puzzle methodically.
    - Remember to take a break every once in a while.
    - Producers may have more than one purpose, keep that in mind when modifying their output.
    - Some puzzles may only have one place that each building fits, use that.
    - Make use of all building variants.
    - One quad cutter is worth 1,000 windmills.
```
And when creating a puzzle: 
```
    - Be careful, making a big puzzle might scare future players.
    - Remember to disable buildings that aren't useful, unless you're trying to be sneaky.
    - Most good puzzles have a "trick" to them, try coming up with one when creating your own
    - Once you finish you puzzle, don't publish right away. Always playtest.
    - Using complex shapes that don't impact the gameplay of your puzzle could be annoying.
    - Misdirecting the player is often the best way to increase difficulty.
    - An easy puzzle can be fun too!
    - If you disable belts, remember to disable splitters as well.
    - Play a few puzzles to get a feel for what makes a puzzle fun.
    - Once you publish, have people playtest and give you feedback.
    - If you are using a lot of blockers, make sure you're not making the solution obvious.
    - Think about what makes your puzzle different, and focus on that.

```
(Needs more)

Let me know if you have any good ideas for more ideas for tips :)


Sense101:
looks cool

PR #1299, opened 2021-09-12
Update base-ro.yaml

Andrei7506238:
- Fixed some issues with previous translation
- Completed the translation of the base game to RO
- Translated the Puzzle DLC to RO

The translation is not always "mot-a-mot". It is rather an adaptation to RO (as some words in EN don't have a direct correspondent in RO). I tried to keep the same formating as before.


tobspr:
Awesome, thanks!

PR #1300, opened 2021-09-17
Added some translation for the main screen lines fo Ru localization

Stubbs943:
Please check and confirm the suggested translation.


Aphanasiy:
Всё-таки "Как настраивать линии эффективнее"

PR #1319, opened 2021-11-22
Button fix when right click with placement building

DJ1TJOO:
When right clicking a button/lever with a placement building it also deletes the button/lever. This PR fixes it by checking if the is a placement building selected.

**Before**
button-fix-before

**After**
button-fix-after


tobspr:
Thanks!

PR #1322, opened 2021-11-25
Remove duplicate hub acceptors

EmeraldBlock:
The hub has duplicate acceptors (left-facing bottom and right-facing bottom). This removes them and reorders the other acceptors to make the positioning clearer.


tobspr:
Already fixed

PR #1325, opened 2021-12-02
use reseed for constructor for rng, they do the exact same thing

mekb-turtle:


mekb-turtle:
unless i'm forgetting something, this shouldn't make any difference apart from making it easier to read i guess

mekb-turtle:
side note: if the seed argument is falsey, such as 0, it will get ignored and use Math.random

EmeraldBlock:
the "random seed" thing is never used, perhaps it's better to change it use a default parameter?

(also, you can type about your PR in the top comment lol, it should be editable)

mekb-turtle:
i've searched in files and i don't see any files containing the word "setSeed" so i assume it's safe to remove

EmeraldBlock:
you can't re-call a constructor function (I guess you could but it'd be insanely hacky), and even then, it's really bad programming practice.

I wasn't referring to the `reseed` function, but to the `|| Math.random()` thing.

mekb-turtle:
yeah, also i forgot to say removed the `Math.random()` thing in the commit name, i did though however as well as removing the `reseed` function

EmeraldBlock:
yeah, I meant possibly replacing it with a default parameter

mekb-turtle:
ah i see, i don't see anywhere it isn't specified so probably no need

EmeraldBlock:
I mean in case it's needed in the future, but I guess it could always be changed then

tobspr:
I still think its useful to have this code, even if its not used right now

PR #1336, opened 2022-01-16
Update JSDoc for Replacing Methods

Bagel03:
This just gives us some autocomplete and type checking when overwriting or extending methods (so you don't have to keep looking at the original)


tobspr:
Nice, thanks!

PR #1337, opened 2022-01-18
Allow external URL to load more than one mod

Bagel03:
Instead of loading the text returned from the remote server, load a JSON object with a `mods` field, containing strings of all the mods. This lets us work on more than one mod at a time or without separate repos. This will break tooling such as `create-shapezio-mod` though.

EDIT: Now you can just set externalmodURL to an array of strings (Which will not beak tooling)

### DIFF
--- a/src/js/mods/modloader.js
+++ b/src/js/mods/modloader.js
@@ -116,8 +116,24 @@ export class ModLoader {
                         response.statusText
                 );
             }
+            
+            let json;
+            try {
+                json = await response.json();
+            } catch (e) {
+                throw new Error(
+                    "Failed to parse JSON object from " +
+                        globalConfig.debug.externalModUrl +
+                        ": " +
+                        e
+                );
+            }
+            
+            if(json.mods === undefined) {
+                throw new Error("Failed to load " + globalConfig.debug.externalModUrl + ": No 'mods' field in json object");
+            }
 
-            mods.push(await response.text());
+            mods = mods.concat(json.mods);
         }
 
         window.$shapez_registerMod = (modClass, meta) => {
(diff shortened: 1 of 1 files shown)


DJ1TJOO:
I do not think this is necessary. Most users will only be developing one mod at a time.

Bagel03:
Maybe not developing but it would help to have 1 repo `<name> shapez mods` rather than a bunch of smaller ones and multiple copies of the game

DJ1TJOO:
You could also use git branches

EmeraldBlock:
would agree that loading multiple mods could be useful (maybe they're related, or you want to test how they interact, or something), but I think JSON's a bit hacky for that

Bagel03:
Yea Emerald, I agree, but I could not think of another way that would be less hacky and I could write in the span of my English class 😄

EmeraldBlock:
so that's why it's a verified commit 🤔

tobspr:
I'd prefer if the externalModUrl could be either string or string[]

Bagel03:
Good idea, I have math next so I'll try to do it then

Bagel03:
Math was a little busy, this should work 👍

src/js/mods/modloader.js
```diff
@@ -103,21 +103,25 @@ export class ModLoader {
...
-                        globalConfig.debug.externalModUrl +
-                        ": " +
-                        response.status +
-                        " " +
-                        response.statusText
-                );
+            let modURLs = Array.isArray(globalConfig.debug.externalModUrl) ? 
+                globalConfig.debug.externalModUrl : [globalConfig.debug.externalModUrl];
+            
+            for(let i = 0; i < modURLs.length; i++) {
```
tobspr, line 109: can you fix the spacing

tobspr (changes requested):

PR #1347, opened 2022-01-21
Replaced concatenated strings with template literals

saile515:
Replaced concatenated strings with template literals for consistency and standard in ES6.


EmeraldBlock:
is the scope of this just the mod examples?

saile515:
> is the scope of this just the mod examples?

Yes.

PR #1348, opened 2022-01-22
Prevent double registering of mods

Bagel03:
Some tools or users may choose to register mods themselves, in which case they should not try to be registered twice


DJ1TJOO:
This should already not be possible as there is a catch when the registerMod fails but if it is called from another place it will just add it.

tobspr:
There is already a check in $shapez_registerMod, what am I missing?

PR #1354, opened 2022-01-24
Add Ability to Extend Constructor Functions

Bagel03:
Unfortunately, you will have to `ClassA = extendConstructor(ClassA, ...)` as there is no way to do that inside of a function


Bagel03:
Whoops

PR #1355, opened 2022-01-24
 Add Ability to Extend Constructor Functions 

Bagel03:
Unfortunately, you will have to `ClassA = extendConstructor(ClassA, ...)` as there is no way to do that inside of a function

### DIFF
--- a/src/js/mods/mod_interface.js
+++ b/src/js/mods/mod_interface.js
@@ -484,6 +484,22 @@ export class ModInterface {
     }
 
     /**
+     * Creates a new class that extends another class (DOES NOT OVERWRITE)
+     *
+     * @template {constructable} C
+     * @param {C} classHandle
+     * @param {bindThis<(...args: ConstructorParameters<C>) => any, InstanceType<C>>} constructor
+     */
+    extendConstructor(classHandle, constructor) {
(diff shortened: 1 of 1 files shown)


EmeraldBlock:
given that it's equivalent to extending a class, why not just extend the class?

Bagel03:
The same could be said about replacemethod no?

EmeraldBlock:
doesn't that modify the original class though?

Bagel03:
Ok, now it modifies it by default, however, it does need a scope to work in (I would like if we could have support for the shapez object so it could be the default parameter for scope)

EmeraldBlock:
oops! I just realized the class constructor can't be modified in place since it's the code of the function lol

honestly I feel like at this point just extending the class and replacing where it occurs feels nicer; also there'd be prototype issues when you do `classWithExtendedConstructor.prototype`

Bagel03:
Would be nice if not for #1356

EmeraldBlock:
but wouldn't overriding `shapez.someClass` only be useful if you instantiated that class yourself, in the mod? Would expect that you'd want to override builtin classes that are called automatically, I guess that might be impossible though

PR #1358, opened 2022-01-28
Fix All Shapez Exports Being Const

Bagel03:
Is this hacky? yes. Is there another way? idk. Either way, this affects bundle size only a tiny amount and just helps modders a lot


tobspr:
Agreed its really hacky, but I think for now its a simple pragmatic solution

PR #1360, opened 2022-02-01
Fix Const Option #2

Bagel03:
This is another option for solving the const problem, which does not use eval but may create more than one function per file


tobspr:
I prefer the first version .. honestly the modloader already uses `eval()`  to load the mods so I prefer the more compact version.

PR #1363, opened 2022-02-01
Refactor acceptor and processor code

Sense101:


Sense101:
this commit history is a mess, will redo nicely

PR #1365, opened 2022-02-01
Update French translations for mods

Bqleine:


tobspr:
Could you keep the same formatting so we can see whats changed? :)

PR #1369, opened 2022-02-03
Modloader custom items fix

DJ1TJOO:
When loading a world with custom items there is no resolver for this item. This will cause the savegame to be corrupted. This PR adds a `registerItem` to the mod interface to easily add the item to the item registery and add a resolver for savegames.


tobspr:
Thanks! This one was actually left on my todo list :)

PR #1370, opened 2022-02-04
Fix that whole export debacle

Bagel03:
I'm 99% confident in this one, but it would be best to do some testing before merging


Bagel03:
This now works (tested both prod standalone and prod browser)

J-Barta:
My friend told me it works (frined is Begal)

tobspr:
Not really nice but I can live with it :p

PR #1371, opened 2022-02-05
Added mod processing requirements

DJ1TJOO:
This mods adds a hook for custom processing requirements

### DIFF
--- a/src/js/game/systems/item_processor.js
+++ b/src/js/game/systems/item_processor.js
@@ -178,7 +178,7 @@ export class ItemProcessorSystem extends GameSystemWithFilter {
         const pinsComp = entity.components.WiredPins;
 
         if (MOD_PROCESSING_REQUIREMENTS[itemProcessorComp.processingRequirement]) {
-            return MOD_PROCESSING_REQUIREMENTS[itemProcessorComp.processingRequirement]({
+            return MOD_PROCESSING_REQUIREMENTS[itemProcessorComp.processingRequirement].bind(this)({
                 entity,
                 item,
                 slotIndex,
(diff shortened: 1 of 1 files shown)


ZygZagGaming:
This PR doesn't modify the `canProcess` method in `item_processor.js`

EmeraldBlock:
> This PR doesn't modify the `canProcess` method in `item_processor.js`

isn't this PR for whether a building can *accept* an item, not whether it's ready to perform processing? (though the naming is confusing)

DJ1TJOO:
> > This PR doesn't modify the `canProcess` method in `item_processor.js`
> 
> isn't this PR for whether a building can _accept_ an item, not whether it's ready to perform processing? (though the naming is confusing)

It is for both. A process can only be done when there are items in the building. If you stop the accepting of the items the process wil also be stopped

EmeraldBlock:
> > > This PR doesn't modify the `canProcess` method in `item_processor.js`
> > 
> > 
> > isn't this PR for whether a building can _accept_ an item, not whether it's ready to perform processing? (though the naming is confusing)
> 
> It is for both. A process can only be done when there are items in the building. If you stop the accepting of the items the process wil also be stopped

yeah, that's what I mean - *directly*, it's only about accepting the item (and it also means it can't do fancier processing requirements)

ZygZagGaming:
What I'm saying is that `canProcess` checks requirements separately and it needs to return true for mod processors to function

EmeraldBlock:
> What I'm saying is that `canProcess` checks requirements separately and it needs to return true for mod processors to function

By default, it simply checks that an item has been accepted in every slot. Filtering items is done when accepting, not processing.

PR #1377, opened 2022-02-06
Update base-fr.yaml for mods translation

Pimak:
Translation for the mods in french


Pimak:
This is duplicate of  #1365 but with a normal formatting

tobspr:
ty!

PR #1379, opened 2022-02-08
added a workaround for mods

ZygZagGaming:
this should allow mod devs to work around the semi-hacky logic for item accepting by registering "filters" in the MOD_ITEM_FILTERS object in item_ejector.js

### DIFF
--- a/src/js/game/systems/item_ejector.js
+++ b/src/js/game/systems/item_ejector.js
@@ -6,6 +6,7 @@ import { StaleAreaDetector } from "../../core/stale_area_detector";
 import { enumDirection, enumDirectionToVector } from "../../core/vector";
 import { BaseItem } from "../base_item";
 import { BeltComponent } from "../components/belt";
+import { inputFilterId } from "../components/input_filter";
 import { ItemAcceptorComponent } from "../components/item_acceptor";
 import { ItemEjectorComponent } from "../components/item_ejector";
 import { Entity } from "../entity";
@@ -13,12 +14,6 @@ import { GameSystemWithFilter } from "../game_system_with_filter";
 import { MapChunkView } from "../map_chunk_view";
 
 const logger = createLogger("systems/ejector");
-
-/**
- * Key is component id, value is function to handle whether or not item can enter.
- * @type {Object<string, (item: BaseItem, reciever: Entity, slot: number) => boolean>}
- */
-export const MOD_ITEM_FILTERS = {};
 
 export class ItemEjectorSystem extends GameSystemWithFilter {
     constructor(root) {
@@ -300,13 +295,10 @@ export class ItemEjectorSystem extends GameSystemWithFilter {
             }
         }
 
-        // it could be a custom mod case
-        // let's check if it is any of them
-        for (let id in MOD_ITEM_FILTERS) {
-            if (receiver.components[id]) {
-                let handler = MOD_ITEM_FILTERS[id];
-                if (handler(item, receiver, slotIndex)) return true;
-            }
+        const inputFilterComp = receiver.components[inputFilterId];
+        if (inputFilterComp) {
+            // it's a custom filter
+            if (inputFilterComp.canAcceptItem(item, receiver, slotIndex)) return true;
         }
 
         return false;
--- a/src/js/game/systems/item_ejector.js
+++ b/src/js/game/systems/item_ejector.js
@@ -14,6 +14,12 @@ import { MapChunkView } from "../map_chunk_view";
 
 const logger = createLogger("systems/ejector");
 
+/**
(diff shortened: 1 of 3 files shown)


src/js/game/systems/item_ejector.js
```diff
@@ -294,7 +300,18 @@ export class ItemEjectorSystem extends GameSystemWithFilter {
...
 
-        return false;
+        // it could be a custom mod case
+        // let's check if it is any of them
+        var any = false; // so that all matching components can process it
+        for (let id in MOD_ITEM_FILTERS) {
+            if (receiver.components[id]) {
+                let handler = MOD_ITEM_FILTERS[id];
+                if (handler(item, receiver, slotIndex)) any = true;
+                else return false;
```
EmeraldBlock, line 310: this makes the system pass over an item if (it would by default) _or_ (_a_ filter passes _and_ _all_ filters pass), which feels like a really awkward combination; particularly, *a* filter increases opportunities, but *more* filters decrease opportunities

src/js/game/systems/item_ejector.js
```diff
@@ -294,7 +300,18 @@ export class ItemEjectorSystem extends GameSystemWithFilter {
             }
         }
 
-        return false;
+        // it could be a custom mod case
+        // let's check if it is any of them
+        var any = false; // so that all matching components can process it
+        for (let id in MOD_ITEM_FILTERS) {
+            if (receiver.components[id]) {
```
EmeraldBlock, line 307: feel like item filters should be handled by a single configurable component, for performance (you already know which filters you need to check); maybe this breaks ECS though

EmeraldBlock (commented):
hard to say much without concrete use cases

src/js/game/systems/item_ejector.js
```diff
@@ -294,7 +300,18 @@ export class ItemEjectorSystem extends GameSystemWithFilter {
             }
         }
 
-        return false;
+        // it could be a custom mod case
+        // let's check if it is any of them
+        var any = false; // so that all matching components can process it
```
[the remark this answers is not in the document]
ZygZagGaming: changed to `let`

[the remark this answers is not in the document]
EmeraldBlock: you appear to have just obliterated it but ok

src/js/game/systems/item_ejector.js
ZygZagGaming: gonna be honest i have no idea what ECS stands for. went ahead and implemented the component anyway. thanks for the suggestion

ZygZagGaming: im like 85% sure it doesn't break ECS now that i know what it means.

src/js/game/systems/item_ejector.js
ZygZagGaming: obliterated too

ZygZagGaming:
> hard to say much without concrete use cases

lets say im making a factory building that takes in shapes, not only shapes, but specifically unpainted one-layer shapes. seems simple but there's currently no non-hacky way to do this with just a mod. better use cases exist, with mods being made for custom item types like liquids, but thats just off the top of my head

EmeraldBlock:
Sorry for just realizing, but
1. does your code actually do anything? you gotta call the appropriate `tryTakeItem` or similar, it seems
2. acceptors themselves have built-in item filters, seems your use case fits that better

ZygZagGaming:
my use case didnt really fit the problem as i intended. i meant more about dynamic input filters, that can change based on, say, a wire input.

ZygZagGaming:
> acceptors themselves have built-in item filters, seems your use case fits that better

the acceptor method doesnt take in a reference to the parent entity, so dynamic usage is difficult.

> does your code actually do anything? you gotta call the appropriate tryTakeItem or similar, it seems

i'm leaving that up to the mod developer. i want as little as possible to be hardcoded.

EmeraldBlock:
> the acceptor method doesnt take in a reference to the parent entity, so dynamic usage is difficult.

you could always just edit it to pass that too

> i'm leaving that up to the mod developer. i want as little as possible to be hardcoded.

I guess so, but then it's not really named well

DJ1TJOO:
What you want can be done with https://github.com/tobspr/shapez.io/pull/1371. It adds a hook for checking if an item is accepted before accepting it. This is the same way the quad painter works doe checking the wires

ZygZagGaming:
> What you want can be done with #1371. It adds a hook for checking if an item needs to be accepted before accepting it. This is the same way the quad painter works for checking the wires

That's a much better way of doing it. Thanks

Sense101:
This is no longer a problem with the latest changes I've made to the refactor... if it ever gets merged :)

PR #1382, opened 2022-02-13
Added a can process hook

DJ1TJOO:
Adds a can process hook (I missed this the first time)


DJ1TJOO:
> could you check the conflicts?

Fixed

PR #1383, opened 2022-02-14
mark all abstract functions abstract

Dimava:
For modding typings purposes, abstract functions should be marked as abstract


tobspr:
ty!

PR #1387, opened 2022-02-17
Translate tips for Japanese

soukouki:
Some of them are in Japanese and English at the same tip, so I fix them too.


tobspr:
Thanks!

PR #1389, opened 2022-02-25
Fix for freeEntityAreaBeforeBuild function.

TcePrepK:
Small typo that brings big bug with it. If you try to replace replaceable entity with wired pins, using unreplaceable entity with wired pins, it crashes because it is looking for "replacing entity" instead of "replaced entity".


Rexogamer:
This appears to have been fixed by 679e7f168cd48463bdcbe43b72bc5cf10eb63897?

PR #1399, opened 2022-03-05
Fix edit dialog not showing when placing a constant producer in the puzzle editor

EmeraldBlock:
During the mod update, the constant signal code was refactored, moving the dialog to choose a signal from `ConstantSignalSystem` to `HUDConstantSignalEdit`. However, the latter is an additional HUD part only added in the regular game mode, meaning the dialog is no longer part of the puzzles editor.

This PR fixes that by adding `HUDConstantSignalEdit` as an additional HUD part in the puzzle editor game mode.


tobspr:
Thanks!

PR #1403, opened 2022-03-13
Update base-ar.yaml

gnarayan1337:


EmeraldBlock:
only translating part of the time translations?

tobspr:
Better than nothing I guess :p

PR #1412, opened 2022-04-04
Simplify colors.js

EmeraldBlock:
This PR simplifies computing `enumColorMixingResults` by using bitwise OR. It also defines `c = enumColors` earlier on to shorten usage more.

### DIFF
--- a/src/js/game/colors.js
+++ b/src/js/game/colors.js
@@ -11,19 +11,20 @@ export const enumColors = {
     white: "white",
     uncolored: "uncolored",
 };
+const c = enumColors;
 
 /** @enum {string} */
 export const enumColorToShortcode = {
-    [enumColors.red]: "r",
-    [enumColors.green]: "g",
-    [enumColors.blue]: "b",
+    [c.red]: "r",
+    [c.green]: "g",
+    [c.blue]: "b",
 
-    [enumColors.yellow]: "y",
-    [enumColors.purple]: "p",
-    [enumColors.cyan]: "c",
+    [c.yellow]: "y",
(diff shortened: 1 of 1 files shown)


tobspr:
Looks great - although it would be nice to have the asserts in? Especially for mods which for example add new colors, I think it's useful to have the mixing of colors checked

dengr1065:
Would it be meaningful to have some asserts that only trigger with --dev? Such mistakes are unlikely to happen while working with the game codebase itself, but may happen when developing mods using outdated examples.

Also, most mods target a version range such as `>=1.5.0` which is incorrect: according to Semantic Versioning, breaking API changes should be done at least in 1.6.0, and mods should target it as `1.6.x` or `~1.6.0`. My personal suggestion is to collect many small refactors and release them as 1.6.0.

Dimava:
> Looks great - although it would be nice to have the asserts in? Especially for mods which for example add new colors, I think it's useful to have the mixing of colors checked

This is not a place to have asserts in - since this just makes the singleton before mods touch it
Where should they be placed? Should it be made with `afterModsloaded` signal or something?

dengr1065:
Such asserts would just run on `gameInitialized`, that's already done to check whether every building has an assigned keybinding.

EmeraldBlock:
what Dimava said; the asserts deleted aren't affected by mods. Anyway, for vanilla, the old asserts made sense since color mixing results had to be manually defined, but I feel the new code is simple enough without such manual definition. Also, maybe mods might do something really bizarre with the color mixing that the asserts wouldn't get?

PR #1426, opened 2022-06-01
Fixes & adding translations

IcedDog:
- Added missing translations.
- Corrected a bunch of mistakes according to "Chinese Copywriting Guidelines".
- Corrected small amount of spelling mistakes.

Also, I noticed there isn't exactly a guideline for keybinds expression while translating, sometimes it uses bold, sometimes it uses quotes, and sometimes it uses nothing. Which type should I go with?


IcedDog:
Updated to v1.5.2

tobspr:
Ty!

PR #1428, opened 2022-06-03
Update CI to use Node.js 16.x

dengr1065:
Right now, CI for linting does not run, as Yarn fails to install dependencies (at least Node.js 12.x is required). In order to fix this, the PR bumps Node.js version to 16.x.


tobspr:
Ty!

PR #1441, opened 2022-06-16
Remove ".io"s from the hungarian translation

Stoneblackdog:
I removed mentions of .io from the hungarian translation


EmeraldBlock:
you can rename the title instead of opening a new PR (#1440)

tobspr:
Thank!

Issue #1444, opened 2022-06-19
The windmill check should use `windmillCount > 2` instead of `windmillCount > 1`


ExE-Boss:
This would allow shapes such as [<code>WuWu?u?u</code>][WuWu?u?u][^1] and [<code><img height="24" align="top" src="https://user-images.githubusercontent.com/3889017/174471636-c4869f34-9444-4734-8f8e-16a8fbe259f6.png" /> WuRuWuRu</code>][WuRuWuRu] to spawn naturally instead of being converted into [<code>RuRu?u?u</code>][RuRu?u?u][^1] and [<code><img height="24" align="top" src="https://user-images.githubusercontent.com/3889017/174471846-e3c1e2df-328e-4286-a466-123bde43495b.png" /> RuRuWuRu</code>][RuRuWuRu] respectively.

## See also
- <https://github.com/tobspr/shapez.io/pull/936>
- <https://github.com/tobspr/shapez.io/pull/1445>

[WuWu?u?u]: https://viewer.shapez.io/?WuWuSuCu
[RuRu?u?u]: https://viewer.shapez.io/?RuRuSuCu
[WuRuWuRu]: https://viewer.shapez.io/?WuRuWuRu
[RuRuWuRu]: https://viewer.shapez.io/?RuRuWuRu
[^1]: `?u` represents any non‑windmill piece

tobspr:
See comment on PR :)

PR #1445, opened 2022-06-19
Fix windmill sub‑shape check in shape generation

ExE-Boss:
* Fixes <https://github.com/tobspr/shapez.io/issues/1444>

--------------------------------------------------------------------------------

This makes it possible for [<code> WuRuWuRu</code>][WuRuWuRu] and other such shapes to spawn naturally.

[WuRuWuRu]: https://viewer.shapez.io/?WuRuWuRu

--------------------------------------------------------------------------------

Supersedes <https://github.com/tobspr/shapez.io/pull/936> in a manner that avoids calling `rng.nextInt()`

### DIFF
--- a/src/js/game/map_chunk.js
+++ b/src/js/game/map_chunk.js
@@ -233,7 +233,8 @@ export class MapChunk {
                 ++windmillCount;
             }
         }
-        if (windmillCount > 1) {
+
+        if (windmillCount > 2) {
             subShapes[0] = enumSubShape.rect;
             subShapes[1] = enumSubShape.rect;
         }


tobspr:
Feel free to make it a mod - but I wouldn't like to merge it into the base game, since it changes the map generation of existing savegames

ExE-Boss:
For the most part, existing save games should be mostly unaffected, except when the shape is far from the Hub and matches `RuRu?u?u`.

--------------------------------------------------------------------------------

One possible workaround would be to make it so that when an extractor is placed on a resource patch, that resource patch gets stored in the save file instead of being re‑generated every time the save is loaded.

--------------------------------------------------------------------------------

Also, making this a mod would require replacing the entire `internalGenerateShapePatch(rng, shapePatchSize, distanceToOriginInChunks)` method.

PR #1448, opened 2022-06-20
Factor out windmill check in shape patch generation

ExE-Boss:
Since <https://github.com/tobspr/shapez.io/issues/1444> (PR <https://github.com/tobspr/shapez.io/pull/1445>) was rejected, this moves that code into a method to allow mods to change how many windmill sub‑shapes are allowed to spawn, and also limit any custom mod sub‑shapes that are allowed to spawn in the world, without replacing the entire `internalGenerateShapePatch(rng, shapePatchSize, distanceToOriginInChunks)` method.

### DIFF
--- a/src/js/game/map_chunk.js
+++ b/src/js/game/map_chunk.js
@@ -226,17 +226,7 @@ export class MapChunk {
             ];
         }
 
-        // Makes sure windmills never spawn as whole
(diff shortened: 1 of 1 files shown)


ExE-Boss:
🔔 @tobspr

PR #1451, opened 2022-06-21
Fix the linting errors 

nightfire2:
Fixes the two small linting errors because it was bugging me. 

### Changes
- **src/js/core/background_resources_loader.js** - removed unnecessary escape character
- **src/js/game/map_view.js** - added eslint override for temporary constant boolean expression


tobspr:
Thanks!

PR #1453, opened 2022-06-21
Update base-es.yaml - Mod translations

DMLON:
Added mod and some missing translations for Spanish. Synced all translations.


tobspr:
Can you make a PR without the syncTranslations? :) Otherwise it will cause a lot of conflict

tobspr:
(Went ahead and just merged your translations)

PR #1466, opened 2022-07-26
Update some string to PT-BT

rogeriocamorim:
For a long time, this was not updated. So here it is.


rogeriocamorim:
@tobspr Why is this faling?

EmeraldBlock:
only **CI / CI** (which checks code) is failing, so it doesn't have anything to do with your PR (it was already failing on the `master` branch)

PR #1485, opened 2022-11-03
Await asynchronous mod initialization

dengr1065:
This change allows mods to delay the game loading process until they're ready. An example of such mod is [Custom CSS/LESS](https://github.com/dengr1065/shapez-mods/tree/master/src/custom_css), which needs to run early and currently relies on the built-in settings system, caching the transpiled CSS.

Use cases:
 - Read custom configuration/settings files before the game loads
 - Wait for third-party/WebAssembly code to load
 - Do other kinds of I/O, such as network calls


tobspr:
Thanks!

PR #1511, opened 2023-03-08
Migrated array of entities in entity_manager.js to map using uid as key

Xiving:
I started running into very slow load/save times in my main world where I built my MAM (~200,000 buildings).

After some investigation using a profile in Chrome, I found that the `registerEntity` function is the culprit of the long load times, more specifically the use of an array to keep track of the entities. The current implementation of `registerEntity` uses `indexOf` to check whether an entity is already registered, which has a quadratic complexity (`O(n^2)`) when registering a substantial amount of entities.

The entity manager already makes use of a UID to distinguish between entity, implements some map-like functionality, such as `findById`, and there is a function that constructs a map which maps UID to entities. All considering, I opted to migrate the list of entities to a map which uses the UID as a key, and replaced some of the aforementioned functionality with their `Map` counterparts.

Theoretically this should already improve loading times, since the complexity of `O(n^2)` is brought down to `O(n log(n))`. But practical results are more fun 😄. I loaded my 200,000 building world up in Chrome (I played on Steam before), and the loading time went down to 10 seconds, whereas right now it takes substantially longer, I tried timing it but after 7 mins in a loading screen I had enough.

### DIFF
--- a/src/js/game/core.js
+++ b/src/js/game/core.js
@@ -440,7 +440,7 @@ export class GameCore {
         this.overlayAlpha = lerp(this.overlayAlpha, desiredOverlayAlpha, 0.25);
 
         // On low performance, skip the fade
-        if (this.root.entityMgr.entities.length > 5000 || this.root.dynamicTickrate.averageFps < 50) {
+        if (this.root.entityMgr.entities.size > 5000 || this.root.dynamicTickrate.averageFps < 50) {
             this.overlayAlpha = desiredOverlayAlpha;
         }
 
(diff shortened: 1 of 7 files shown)


Bagel03:
Hey, and thanks for looking into this. Unfortunately, it doesn't look like the base game will be getting any more updates, however there is a [community fork](https://github.com/tobspr-games/shapez-community-edition) that is currently being worked on before it is released. If you would like to you can open another issue or pr there and more people can discuss it.

PR #1533, opened 2023-04-13
Update base-fr.yaml

Didydrogene:
Translated : 
browserNoSupport (1427)
mods (1146) - This one is kinda weird, as the capital letter makes no sense to me (I'm not a native english-speaker)

Corrected :
description (1423) - This is purely objective, but I do think it sounds better with the change.
modsInfo (1416) - Depends on whether you wish to adress directly to the user or speak in a more general way. Although this general time is not used anywhere else as is.
desc (918) - Fixed an inconsistency in the way the wire layer is called
desc (856) - Sneaky double space right there
desc (841) - Lil comma lacking compared to the desc of the merger, whiwh has the same exact structure.
desc (802) - Lacking a space after the comma. Cutting in half that very long sentence makes it easier to read quickly. I'll come back to this description because it's a bit rough and there's some strange stuff

Proposed :
desc (862) - "barrière" is way more common in french than "drain" and makes it easier to understand, at least to me. I may be influenced by the building with the same features in the game Mindustry, which is translated as "barrière de débordement".

### DIFF
--- a/translations/base-fr.yaml
+++ b/translations/base-fr.yaml
@@ -800,9 +800,9 @@ storyRewards:
     reward_stacker:
         title: Assembleur
         desc: Vous pouvez maintenant assembler deux formes avec le
-            <strong>combineur</strong> ! Les deux entrées sont assemblées,si
+            <strong>combineur</strong> ! Les deux entrées sont assemblées ; si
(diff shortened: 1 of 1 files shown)


Didydrogene:
First time on github, if there's anything wrong with my process of doing things please tell me !

PR #1548, opened 2023-04-28
Remove emphasis in Cutter in `base-en.yaml`

KonkenBonken:
Removed `strong` tags as they aren't necessary.
Here, it emphasizes the warning as if it was a very important warning which I believe is not true.
If you go against the warning, you'll lose some temporary efficiency, but that's all. It's just part of learning the game, 
hence not needing to be in a strong font

You might also want to remove the exclamation mark or even the whole warning.

### DIFF
--- a/translations/base-en.yaml
+++ b/translations/base-en.yaml
@@ -835,10 +835,10 @@ buildings:
     cutter:
         default:
             name: &cutter Cutter
-            description: Cuts shapes from top to bottom and outputs both halves. <strong>If you use only one part, be sure to destroy the other part or it will clog and stall!</strong>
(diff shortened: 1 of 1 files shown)


EmeraldBlock:
Personally, I think it is actually an important warning: Stalling just outright blocks your factory, and it's not immediately obvious why a cutter stops. (Occasionally I see someone who uses the outputs in an imbalanced way, so one path backs up, but they are still confused why it stops.) Anyway, the tone is more informal and conversational, so such emphasis and an exclamation mark is not necessarily inappropriate.

Also the way translations are synced treats tags as placeholders as well, i.e. for a given translation, the tags used all match.

PR #1555, opened 2023-06-07
Update base-ar.yaml

mhd-fettah:
adding Arabic translation .


mhd-fettah:
I will continue with the rest since I am a native speaker, and I have significant experience in PHP Laravel. I have also done a lot of localization work for various projects, including business and games. Additionally, I am a player of the game, so I understand the exact context of each sentence.

PR #1637, opened 2026-01-10
Updated ukrainian translate.

LanternOfDarkness:
Added Ukrainian translation for missing strings.


LanternOfDarkness:
I signed the agreement, but the check is still pending, even after refreshing.

